PR #3, opened 2017-11-01
Second wave of iele changes

dwightguth:
Changes to IELE itself:

* Change CALL* and LOCALCALL opcodes to accept a list of register arguments and a list of register return values
    * It is an exception if the number of returned values does not match the number of return registers in the call instruction
    * CALL, STATICCALL, DELEGATECALL, CALLCODE, and LOCALCALL take 2 16-bit immediate values representing the number of arguments and return registers.
    * RETURN and REVERT take 1 16-bit immediate value representing the number of values to return.
    * the CALLDATA* instructions are removed; arguments to the call are loaded into registers 0 to n-1 of the callee.
    * the RETURNDATA* instructions are removed; a later PR will change the signature of CREATE to return failure data directly.
    * Precompiled contracts are modified to accept some fixed number of registers representing their real arguments, as opposed to as a single memory buffer
* the LOCALRETURN and RETURN instructions are merged: RETURN returns from LOCALCALL if in a local call, otherwise it returns from the entire contract
* an MLOAD and MSTORE instruction are added for fixed-but-arbitrary width load/store
   * each instruction takes an additional operand N, a width in bytes, and interprests the bytes as 2's complement modulo 256^N

Major other changes:

* deleted .k files for now; the repo is not ready for examination by the public, so we don't need to worry about not having pandoc-tangle
* removed references to EVM from text
* substantial improvements to EVM-to-IELE conversion
* remove all outdated EVM schedules


dwightguth:
I'm well aware that the EVM-to-IELE code is kind of a mess. I'm happy to work with whoever picks up that part of the project to help them improve it, but I think it would benefit from more specific feedback so I know what parts of the code are harder to understand, so I can focus effort there. I tried to address all of your comments, but if there's more changes you'd like to see, let me know.

dwightguth:
Wikipedia has a list of epithets used to describe various iele in Romanian mythology. It seemed like a decent placeholder.

PR #11, opened 2017-11-14
Fix a race between the #newaccount and #exception branches of …

virgil-serbanuta:
…#transferfunds.


virgil-serbanuta:
All tests seemed to pass.

PR #12, opened 2017-11-21
Fix syntactic ambiguity

virgil-serbanuta:
Without this change
log @a,@b,@c,@d=call @d()
could be parse as either

log @a,@b,
@c,@d=call @d()

or

log @a,@b,@c,@d
=call @d()

### DIFF
--- a/iele-syntax/iele-syntax.k
+++ b/iele-syntax/iele-syntax.k
@@ -16,11 +16,13 @@ module IELE-SYNTAX
   // Operands
   syntax LValue ::= LocalName | GlobalName
   
-  syntax LValues ::= List{LValue, ","} [klabel(lvalueList)]
+  syntax NonEmptyLValues ::= LValue | LValue "," NonEmptyLValues
 
   syntax Operand ::= LValue | Constant
 
   syntax Operands ::= List{Operand, ","} [klabel(operandList)]
+
+  syntax NonEmptyOperands ::= Operand | Operand "," NonEmptyOperands
 
   // Assignment
   syntax AssignInst ::= LValue "=" Operand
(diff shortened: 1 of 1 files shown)


theo25:
I don't understand the problem here. Either parsing results in an invalid iele program. It doesn't matter what the parser chooses, it will produce a parsing error, which is the correct behavior.

virgil-serbanuta:
My knowledge of k is still limited, but I assumed that the Operands and LValues lists can be empty. If that is not the case, then I will drop this pull request.

If they can be empty then log `@A,@b,@c,@d=call @e()` is a valid sequence of instructions which can be parsed in two ways. A full contract would be:
```define @f(){log @A,@b,@c,@d=call @e()}```

**Comment by theo25** [comment]
Never mind, you are right. I see the problem now. But I think the root of the problem is that the grammar as is right now allows this as valid iele:

log @A,@b,

The grammar should not allow that. It should be something like:

syntax LogInst ::= "log" /* index */ Operand "," /* size in bytes */ Operand
syntax LogInst ::= "log" /* index */ Operand "," /* size in bytes */ Operand "," NonEmptyOperands
syntax NonEmptyOperands ::= Operand | Operand "," NonEmptyOperands

Does this seem to solve the problem?

**Comment by theo25** [comment]
Same for LValues. It should not be allowed to be empty.

**Comment by virgil-serbanuta** [comment]
Does it look reasonable now?

File: iele-syntax/iele-syntax.k [hunk 1]
```diff
@@ -64,15 +66,18 @@ module IELE-SYNTAX
...
   syntax JumpInst ::= "br" IeleName
   syntax CondJumpInst ::= "br" Operand "," IeleName
-  syntax LocalCallInst ::= LValues "=" "call" GlobalName "(" Operands ")"
-  syntax AccountCallInst ::= LValues "=" "call" GlobalName "at" Operand "(" Operands ")" "send" Operand "," "gaslimit" Operand
+  syntax LocalCallInst ::= "call" GlobalName "(" Operands ")"
+                         | NonEmptyLValues "=" "call" GlobalName "(" Operands ")"
+  syntax AccountCallInst ::= "call" GlobalName "at" Operand "(" Operands ")" "send" Operand "," "gaslimit" Operand
+                           | NonEmptyLValues "=" "call" GlobalName "at" Operand "(" Operands ")" "send" Operand "," "gaslimit" Operand
   syntax SendInst ::= "send" /* value */ Operand "to" /* account */ Operand
   syntax ReturnInst ::= "ret" Operands | "ret" "void"
```
theo25: Should also be NonEmptyOperands

iele-syntax/iele-syntax.k
```diff
@@ -64,15 +66,18 @@ module IELE-SYNTAX
...
   syntax RevertInst ::= "revert" Operands | "revert" "void"
```
theo25: Should also be NonEmptyOperands

virgil-serbanuta: Done.

iele-syntax/iele-syntax.k
virgil-serbanuta: Done.

PR #13, opened 2017-11-22
Iele parser.

virgil-serbanuta:
Not fully optimized because Iele is still being developed and further optimizations would make the parser much harder to change.

Parses Theo's syntax.


bmmoore:
This seems to be missing a `IeleParserImplementation.hs`

virgil-serbanuta:
I think that the "Use applicative operators" commit has some good changes (e.g. parens, braces), but, to me, some seem a step back from the original code. I think that code should be optimized for readability, not for writability, speed, or other things unless there is a pressing need, and I think that in many cases it takes a longer time to read and understand the new code (although it is shorter).

I may be wrong, of course (and please tell me if that's the case), but here are some examples:
```
binaryOperationTypes :: Parser BinaryOperationType
binaryOperationTypes =
      try (enumParser "add" AddInst)
  <|> try (enumParser "mul" MulInst)
  <|> try (enumParser "sub" SubInst)
  <|> try (enumParser "div" DivInst)
  <|> try (enumParser "exp" ExpInst)
  <|> try (enumParser "mod" ModInst)
  <|> try (enumParser "and" AndInst)
  <|> try (enumParser "or" OrInst)
  <|> try (enumParser "xor" XorInst)
  <|> try (enumParser "sha3" Sha3Inst)
  <|> try (enumParser "byte" ByteInst)
  <|> try (enumParser "sext" SExtInst)
  <|> enumParser "twos" TwosInst

binaryOperationInst :: Parser BinaryOperationInst
binaryOperationInst = do
    left <- lValue
    void equal
    operationType <- binaryOperationTypes		
    first <- operand		
    void comma		
    second <- operand		
    return (BinaryOperationInst operationType left first second)
```
vs
```
binaryOperations :: [(String,BinaryOperationType)]
binaryOperations =
  [("add",AddInst)
  ,("mul",MulInst)
  ,("sub",SubInst)
  ,("div",DivInst)
  ,("exp",ExpInst)
  ,("mod",ModInst)
  ,("and",AndInst)
  ,("or", OrInst)
  ,("xor",XorInst)
  ,("sha3",Sha3Inst)
  ,("byte",ByteInst)
  ,("sext",SExtInst)
  ,("twos",TwosInst)
  ]

binaryOperationInst :: Parser BinaryOperationInst
binaryOperationInst =
  flip BinaryOperationInst <$> lValue <* equal <*> enumParser binaryOperations
```

I agree that the original version can be improved.

In the second example, one thing to notice is that human communication is redundant, and that redundancy is essential for efficient communication and for decreasing the mental processing load for the listener/reader (of course, there is a trade-off related to message size and processing load). As such, using field names when constructing an object is a very useful technique for this purpose:

```
loadInst = do
    left <- lValue
    void equal		
    skipKeyword "load"		
    index <- operand		
    void comma		
    sizeInBytes <- operand		
    return LoadInst {loadLeft = left, loadIndex = index, loadSizeInBytes = sizeInBytes}
```
vs
```
loadInst =
  LoadInst <$> lValue <* equal <* skipKeyword "load" <*> operand <* comma <*> operand
```

Note that to actually understand the second example, as opposed to just glossing over it, one needs to be able to mentally group things as:
```
LoadInst <$> (
  (lValue <* equal <* (skipKeyword "load"))
  <*> (operand <* comma)
  <*> operand
)
```
which, maybe, a seasoned Haskell programmer would be able to do at a glance (though is it really so?), but then one must also be able to map the first LoadInst argument to `loadLeft`, the second to loadIndex and the third to loadSizeInBytes, which needs attention shift between the formula which one is reading and the definition of LoadInst. Right now, this definition is close to loadInst, which helps but does not solve the problem, but this may not be the case if we decide to split the AST representation from parsing, which is likely to be needed if we want to increase the parsing speed.

Also, using field names reduces the chance of errors when refactoring or changing the definition of LoadInst, and a smaller chance of errors is always a very useful thing to have.

And I admit that I enjoy writing short and efficient code, and I do that in many of my personal projects.

bmmoore:
This change is aimed for readability. Compare the revised definitions to a conventional grammar definition like

```syntax BinaryOperation ::= LValue "=" BinaryOperationType Operand "," Operand
syntax LoadInst ::= LValue "=" "load" Operand "," Operand```

Your grouping of the production is not necessary or even well-typed. The actual operator precedence is simply

`(((((LoadInst <$> lValue) <* equal) <* skipKeyword "load") <*> operand) <* comma) <*> operand`

virgil-serbanuta:
While I agree that the entire commit is aimed for readability, and that in many cases it succeeds, I wholeheartedly disagree that it succeeds in all cases, these two examples included.

My previous experience with such discussions is that the chance of something useful happening are (very) low after the first exchange, especially if the discussion is not face to face. As such, I did not bring any argument now, I'm just making sure that I'm not being misunderstood as either disagreeing with the entire commit or being convinced by your reply (which is technically correct, but misses the point I was trying to make).

PR #17, opened 2017-11-23
Add memory costs for 'LOCALCALL'.

virgil-serbanuta:


gas.md
```diff
@@ -381,6 +381,8 @@ We assume that all operations interrogating the local state have complexity
   ```hs
   computationCost(LOCALCALL(_, nARGS, _) _ rARGS) =
     saveContextCost + wordCopyCost * (sum [registerSize r | r <- rARGS]) + jumpCost
+  requiredRegisterMemory(LOCALCALL(_, nARGS, _) _ rARGS) =
+    currentRegisterMemory + sum [registerSize(r) | r <- rVALUES]
```
traiansf: One question we should answer is where do we give the extra memory allowance for the call.

virgil-serbanuta: Is there a memory allowance for local calls, or only for calls?

virgil-serbanuta: Since this is still an open question, I added a TODO.

virgil-serbanuta: After a discussion with Brandon it seems that there is a memory allowance for local calls (and local calls only), so I added a flat memory allowance per stack level, and I added some notes about other options.

bmmoore:
There is not supposed to be any extra free memory for making local calls!
Calls into another contract run in a new isolated VM state, so they have separate gas accouning, including calculating the cost of their (separate) memory consumption against a (separate) initial allowance. I would not call this an "additional" allowance because it is not an allowance applied to the memory used in the caller. The only connection between gas calculations in different inter-contract calls should be in the mechanism we use to limit inter-contract call depth.

The only "allowance" is that the gas cost for (local) memory consumption is calculated only from memory usage above a certain constant, which equals the maximum size (in bytes) of the EVM stack. This is done because we need to consider the size of registers as memory consumption to avoid abusing arbitrarily large values - both registers in active use and register values saved in local calls - , but we don't want contracts using registers as a reasonable replacement for stack to have a more complicated gas model than on the EVM, where the stack has a hard depth limit but no incremental cost for using more stack. I think we plan to just include registers in a calculation of local memory and apply an initial allowance to the total figure, but it might also be reasonable to apply the allowance only to the space used by registers and the (local) call stack.

virgil-serbanuta:
Ok, I got confused by the previous discussions on this topic. I'm glad to get a clear answer, I fixed the pull request.

FWIW:
1. I would apply the allowance to memory+registers because they are actually the same thing. The counterargument would be that optimizing for keeping data in (local) registers as opposed to memory may lead to less bugs (since memory is composed of global shared variables).
2. Resetting memory usage for inter-contract calls would lead people that need a lot of memory to split their code between multiple contracts whenever possible to avoid quadratic costs (and benefit from the free EVM stack memory), so it will be interesting to see how that mechanism to limit inter-contract calls will turn out to be (i.e. if we care about the cases where one can split memory usage across contracts, then presumably a call will have a quadratic cost in the depth of the inter-contract stack, otherwise there may be no point in having quadratic memory costs within a contract).

gas.md
```diff
@@ -380,7 +380,10 @@ We assume that all operations interrogating the local state have complexity
   the call site.
   ```hs
   computationCost(LOCALCALL(_, nARGS, _) _ rARGS) =
-    saveContextCost + wordCopyCost * (sum [registerSize r | r <- rARGS]) + jumpCost
+    saveContextCost + wordCopyCost * (sum [registerSize r | r <- rARGS])
+    + jumpCost + callStackDepthCheckCost
+  requiredRegisterMemory(LOCALCALL(_, nARGS, _) _ rARGS) =
+    currentRegisterMemory + sum [registerSize(r) | r <- rVALUES]
```
bmmoore: 1. The memory cost needs to include all the space to save the caller's registers.

2. rVALUES is not defined. Should it be rARGS like in the computation cost.

virgil-serbanuta: Done. Could you take another look?

gas.md
```diff
@@ -378,14 +378,30 @@ We assume that all operations interrogating the local state have complexity
   Save the local context (including the current register stack memory
   requirements), copy the arguments into the new context and jump to
   the call site.
+
+  We're assuming that, in an efficient implementation, saving the caller's
+  registers is a small constant time operation. One way or another, an
```
bmmoore: This sounds like there is a single constant cost to save the entire collection of the caller's registers. I think we should only assume that there is a constant cost for each register (which is what the formula below says).

PR #22, opened 2017-11-29
IELE textual syntax.

theo25:
This PR introduces the textual syntax of IELE and modifies the semantics rules to operate on text encoding instead of binary.


iele-binary.md
```diff
@@ -5,104 +5,300 @@ Here we define an ad-hoc binary encoding for IELE. This encoding is subject to c
...
+                    | BinOp
+                    | FiveOp
+
+    syntax Contract ::= #dasmContract ( WordStack , IeleName )       [function]
+                      | #dasmContract ( WordStack , Int , Map, IeleName , TopLevelDefinitions, Int , Int ) [function, klabel(#dasmContractAux)]
+ // -----------------------------------------------------------------------------
+    rule #dasmContract( .WordStack, _)      => #emptyCode
+    rule #dasmContract( 99 : NBITS : WS, NAME ) => #dasmContract(WS, NBITS, 0 |-> "init", NAME, .TopLevelDefinitions, 1, #sizeWordStack(WS) +Int 2)
+    rule #dasmContract( 105 : W1 : W2 : WS, NBITS, FUNCS, NAME, DEFS, N, SIZE ) => #dasmContract(#drop(W1 *Int 256 +Int W2, WS), NBITS, N |-> #parseToken("IeleName", #unparseByteStack(#take(W1 *Int 256 +Int W2, WS))) FUNCS, NAME, DEFS, N +Int 1, SIZE )
+    rule #dasmContract( 106 : W1 : W2 : WS, NBITS, FUNCS, NAME, DEFS, N, SIZE ) => #dasmContract(#take(W1 *Int 256 +Int W2, WS), N) ++Contract #dasmContract(#drop(W1 *Int 256 +Int W2, WS), NBITS, FUNCS, NAME, external contract N DEFS, N +Int 1, SIZE)
```
bmmoore: This decoding of external contracts looks a bit strange. It looks like the main contract ends up at the end of a list which flattens the child contracts and any contracts nested into those, while the parent contract records the reference as a declaration just storing `N`, which seems to be an index into the parent contract's constant table that makes no sense as an offset into the flattened list of contracts.

theo25: I pushed a fix that uses prefixes as discussed.

iele-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+  syntax UnlabeledBlock ::= Instructions
+
+  syntax Blocks [flatPredicate]
+  syntax Blocks ::= UnlabeledBlock LabeledBlocks | LabeledBlocks
+
+  // Functions
+  syntax FunctionSignature ::= GlobalName "(" FunctionParameters ")"
+
+  syntax FunctionParameters ::= LocalNames
+                              | Int /* when desugared to just the number of parameters */
```
grosu: This should go into the module where the desugaring happens.  We want the IELE-SYNTAX module to only hold the syntax that can be used for parsing IELE programs.

theo25: Done!

iele.md
```diff
@@ -1439,11 +1267,16 @@ For each `CALL*` operation, we make a corresponding call to `#call` and a state-
          </account>
          <activeAccounts> ... ACCTTO |-> (EMPTY => false) ... </activeAccounts>
 
-    syntax KItem ::= "#codeDeposit" Int Int Ops Reg
-                   | "#mkCodeDeposit" Int Int Ops Reg
-                   | "#finishCodeDeposit" Int Ops Reg
+    syntax Contract ::= #subcontract ( Contract , IeleName ) [function]
  // -------------------------------------------------------
-    rule <k> #exception ~> #codeDeposit _ _ _ REG => #popCallStack ~> #popWorldState ~> #popSubstate ~> #load REG 0 ... </k> <output> _ => .Regs </output>
+    rule #subcontract ( (contract NAME ! _ { _ } #as CONTRACT) _, NAME ) => CONTRACT
```
bmmoore: This needs to be consistent with the decoding of child contracts.

iele-binary.md
```diff
@@ -5,104 +5,302 @@ Here we define an ad-hoc binary encoding for IELE. This encoding is subject to c
...
+                    | BinOp
+                    | FiveOp
+
+    syntax Contract ::= #dasmContract ( WordStack , IeleName )       [function]
+                      | #dasmContract ( WordStack , Int , Map, IeleName , TopLevelDefinitions, Int , Int ) [function, klabel(#dasmContractAux)]
+ // -----------------------------------------------------------------------------
+    rule #dasmContract( .WordStack, _) => #emptyCode
+    rule #dasmContract( 99 : NBITS : WS, NAME ) => #dasmContract(WS, NBITS, 0 |-> "init", NAME, .TopLevelDefinitions, 1, #sizeWordStack(WS) +Int 2)
+    rule #dasmContract( 105 : W1 : W2 : WS, NBITS, FUNCS, NAME, DEFS, N, SIZE ) => #dasmContract(#drop(W1 *Int 256 +Int W2, WS), NBITS, N |-> #parseToken("IeleName", #unparseByteStack(#take(W1 *Int 256 +Int W2, WS))) FUNCS, NAME, DEFS, N +Int 1, SIZE )
+    rule #dasmContract( 106 : W1 : W2 : WS, NBITS, FUNCS, NAME, DEFS, N, SIZE ) => #dasmContract(#take(W1 *Int 256 +Int W2, WS), NAME ++IeleName N) ++Contract #dasmContract(#drop(W1 *Int 256 +Int W2, WS), NBITS, FUNCS, NAME, external contract N DEFS, N +Int 1, SIZE)
```
bmmoore: This gives the child contract the name `NAME ++IeleName N` but gives the parent contract a reference to name `N`. That won't work with the current implementation of the create call.

iele-binary.md
```diff
@@ -5,104 +5,302 @@ Here we define an ad-hoc binary encoding for IELE. This encoding is subject to c
...
+    rule #dasmContract( WS, NBITS, FUNCS, NAME, DEFS, N, SIZE ) => contract NAME ! SIZE { DEFS ++TopLevelDefinitions #dasmFunctions(WS, NBITS, FUNCS) } .Contract [owise]
+
+    syntax Contract ::= Contract "++Contract" Contract [function, klabel(contractAppend)]
+    syntax priorities contractDefinitionList > contractAppend
+    syntax TopLevelDefinitions ::= TopLevelDefinitions "++TopLevelDefinitions" TopLevelDefinitions [function, klabel(topLevelAppend)]
+    syntax priorities topLevelDefinitionList > topLevelAppend
+    syntax IeleName ::= IeleName "++IeleName" IeleName [function]
```
bmmoore: The name  `++IeleName` is very misleading - I'm astounded that it does anything but concatenate strings parts of the names. Change it to something like `extendDottedName` or  `+.+IeleName` or `++IeleNamePath`, or anything that doesn't look like unpunctuated concatenation.

PR #23, opened 2017-11-30
improve documentation and fix minor issues

dwightguth:
This fixes the following things in the semantics:

* Adding code to disassemble global registers
* Fix bug with arity of iele.ecpairing
* Add case for copycreate of empty account
* Rename coinbase to beneficiary
* Throw an exception when dividing by zero
* Remove gas cost for copying existing contracts
* Remove callcode and delegatecall
* Arbitrary precision storage keys

I also add significant documentation to the semantics and clean up references to ethereum and unused files.


tests/evm-to-iele/conversion.ml
```diff
@@ -784,7 +784,7 @@ let alloc_registers (ops: iele_op list) : iele_op list =
     regbits := !regbits + 1;
     regcount := !regcount asr 1
   done;
-  VoidOp(`REGISTERS !regbits,[]) :: VoidOp(`FUNCTION("deposit"),[]) :: VoidOp(`FUNCTION("iele.ecrec"),[]) :: VoidOp(`FUNCTION("iele.id"),[]) :: VoidOp(`FUNCTION("iele.ecadd"),[]) :: VoidOp(`FUNCTION("iele.ecmul"),[]) :: VoidOp(`CALLDEST(0, 0), []) :: (lbl_ops @ dangling_jumpdests) 
+  VoidOp(`REGISTERS(!regbits + 1),[]) :: VoidOp(`FUNCTION("deposit"),[]) :: VoidOp(`FUNCTION("iele.ecrec"),[]) :: VoidOp(`FUNCTION("iele.id"),[]) :: VoidOp(`FUNCTION("iele.ecadd"),[]) :: VoidOp(`FUNCTION("iele.ecmul"),[]) :: VoidOp(`CALLDEST(0, 0), []) :: (lbl_ops @ dangling_jumpdests)
```
bmmoore: Why does this need the `+ 1`?

dwightguth: because we aren't using global registers so we need an extra bit to ensure that the local registers don't become global

PR #27, opened 2017-12-01
Documentation for textual IELE

theo25:


Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
-* Registers are divided into two categories: global registers and local registers. Half of the declared registers are global; the other half are local.
+* Because the stack no longer exists, the POP, DUP, SWAP, and PUSH instructions no longer exist. They are replaced with an assignment instruction that can be used to load a constant value into a register (see section on arbitrary-precision integers for more details), and/or copy a value from one register to another.
+* Registers are divided into two categories: global registers and local registers. Local registers are visible only within the function in which they are used, while global registers are visible across all functions of a given contract. Similar to LLVM, global registers names are prefixed by `@`, while local register names are prefixed by `%`.
+* Local registers have a lifetime equal to the lifetime of a function call to the function that contains them, while global registers have a lifetime that starts with the creation of an account with an instance of the contract that contains them and ends with the destruction of said account.
```
grosu: What happens if a register is used before it is assigned any value?  Are they initialized with 0 by default?  If yes, we should say so.  That also means that if the same register is used in multiple functions, for example `%1`, then it will be `0` initially in each function, no matter if others write it.  Something along this line could be good to mention.  In the Viper project, we found that people in this field are not that trained in PL.

bmmoore: This sounds like global register values are preserved between transactions, like storage.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* IELE has a `call .. at` and a `staticcall .. at` instruction used for calls to public functions of other contracts (the target account address is a register operand of the instruction, as is the amount of gas to be spent during the call, and the value to send in case of `call .. at`). `call .. at` and `staticcall .. at` take an arbitrary number of arguments and return values in the form of register operands instead of memory ranges. The arguments are copied into local registers corresponding to the formal arguments of the function being called. When the contract returns, the returned values are copied into the return value registers of the call instruction.
+* If the name of the function being called does not correspond to a public function of the contract being called, an exception occurs.
+* In all cases, if a mismatch occurs between the number of arguments or return values of a function and the matching `call` or `ret` instruction, an exception occurs.
+* Because they are no longer needed, we remove the CALLDATA* instructions and the RETURNDATA* instructions.
+* Because of security concerns, we remove the CALLCODE and DELEGATECALL instructions.
```
bmmoore: Maybe mention we plan to provide a different way to cheaply reuse code between accounts?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* Registers are divided into two categories: global registers and local registers. Local registers are visible only within the function in which they are used, while global registers are visible across all functions of a given contract. Similar to LLVM, global registers names are prefixed by `@`, while local register names are prefixed by `%`.
+* Local registers have a lifetime equal to the lifetime of a function call to the function that contains them, while global registers have a lifetime that starts with the creation of an account with an instance of the contract that contains them and ends with the destruction of said account.
 
 ## Program Structure
 
-Unlike EVM, in which a contract is a sequence of instructions, a IELE contract consists of a header, followed by a data segment, followed by one or more functions.
+Unlike EVM, in which a contract is a sequence of instructions, a IELE contract consists of a header giving the contract a name, followed by one or more function definitions and optionally one or more external contract declarations.
```
bmmoore: You need to previously mention that the program structure is a list of earlier contracts, otherwise only saying that the declaration refers to an "external contract" by just giving the name doesn't sound like we are packaging child contracts into the parent contract.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
-* Because the function being called is now explicitly part of the signature of calling a contract, we condense the 7 remaining precompiled contracts into a single contract with seven functions.
+* IELE has a `call` instruction used for local calls to other functions within the contract. `call` takes a function name, an arbitrary number of register arguments, and an arbitrary number of register return values, and jumps to the start of the function in the current contract with the corresponding name. It pushes the current instruction position, the values of all local registers, and the register operands of the return values onto the local call stack. The arguments of the call are copied into local registers corresponding to the formal arguments of the function being called.
+* The `ret` and `revert` instructions take an arbitrary number of register operands. `revert` reverts the contract, but `ret` will return to the instruction position on the top of the local call stack if one exists, restoring the values of local registers and copying the return values into the return value operands of the call instruction. If the local call stack is empty, it returns from the contract.
```
bmmoore: How does the caller receive the values if revert is called with multiple arguments?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* A function definition includes the function signature, the function body and whether or not the function is public. A function signature includes a function name (which is also represented as a global register, as it is visible across the contract) and names of formal arguments (which are represented as local registers, as they are only visible within the function body).
+* A public function can be called by other accounts, while a non-public one can only be called by other functions within the same contract.
+* A special function named `@init` is automatically added to the contract (and can be overridden by the programmer), and will be called when an account is created with this contract.
+* An account to which code has never been deployed contains an implicit public function `@deposit` which takes no arguments, returns no values, and does nothing. This function exists to allow accounts to receive payment even if they do not have a contract deployed to them. Note that a contract can forbid payments by refusing to declare the `@deposit` function, and explicitly raising an exception if any of its entry points are invoked with a balance transfer.
```
grosu: I do not understand this.  If the deposit function does nothing, then how can the account receive payment?  Also, looks questionable to allow either a function or a contract in an account.  Or both?  Or a contract with a deposit function in it?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* The contract is malformed if it contains multiple function definitions with the same name.
 
 ## Static Jumps
 
-IELE no longer has dynamic jump instructions. They are replaced with jumps that take an immediate value as an argument, and with a call/return instruction system to allow function calls that return to the caller.
+IELE no longer has dynamic jump instructions. They are replaced with jumps that take a named label as an argument, and with a call/return instruction system to allow function calls that return to the caller.
```
grosu: `instruction system` ?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
 
-Unlike EVM, for ease of verification we disallow creation of dynamic code for dynamically created contracts, and instead restrict dynamically-created-contracts to code that consists of some previously validated contract.
+* IELE has one integer comparison instruction, namely `cmp`, that accepts various predicates. In addition to the predicates `lt`, `gt`, and `eq`, which have corresponding EVM instructions, the predicates `le`, `ge`, and `ne` are also supported.
+* Various EVM instructions that query local and/or network state (e.g. GAS, BALANCE, etc.) have been replaced by IELE builtins that can be called using the `call` instruction (e.g. `%balance = call @iele.balance(%bank.account)`). The names of all IELE builtins start with the prefix `@iele.`, following a convention similar to the LLVM intrinsics. All names for globals (registers and/or functions) starting with this prefix are reserved: A contract is malformed if it contains user-defined registers and/or functions with names starting with `@iele.`.
```
bmmoore: I'm not sure we should say the instructions are "replaced" - the syntax is now written as a call to special names, but we do still have opcodes. Maybe we could have more generic language about providing a uniform syntax as builtin operations?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* Precompiled contracts in IELE are also invoked using corresponding builtins (e.g. `@iele.ecpairing`).
```
bmmoore: Are the precompiled contracts accessed as a local call, or do they need to be written as a remote call that passes an appropriate amount of gas?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+## Contract Creation
 
-## Code Size
+Unlike EVM, both for ease of verification and due to security concerns, we disallow creation of dynamic code for dynamically created contracts, and instead restrict dynamically-created-contracts to code that consists of some previously validated contract.
 
-The maximum length of a IELE contract is expanded to 65536 bytes.
+* The `create` instruction is modified from its EVM counterpart. Instead of taking a memory start and a memory width containing initialization code, the `create` instruction takes a contract name, and an arbitrary number of register arguments to be passed to the `@init` function of the contract to be created. If the contract name is not an externally declared contract name, the contract is malformed.
+* Since each externally declared contract should have been defined in the same file and above the contract that externally declares it, the exact code that will be deployed by a `create` instruction is statically known and available.
```
bmmoore: This is the first place you mention that an "external contract name" actually has to refer to something written in the same file from which the contract code was compiled.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
-* A jump table is computed for each function mapping 16-bit jump labels to the instruction position of the JUMPDEST instruction which declares that label. The contract is malformed if multiple JUMPDEST instructions in the same function have the same label.
-* The JUMP and JUMPI instructions jump to the instruction positon in the jump table for their respective jump label.
-* Instead of throwing an exception if the argument of JUMP or JUMPI is not the program counter of a JUMPDEST instruction, the contract is malformed if the immediate argument of the JUMP or JUMPI instruction is not in the jump table for the function containing the jump.
+* The code within a function contains named labels. Any instruction can be prefixed by a named label that makes this instruction a valid target of a jump.
+* The `br` instruction has two variants. It can used with a named label as a single argument for unconditional jumps within a function body, or with an additional register holding a condition value for conditional jumps when the condition value is not zero. The instruction prefixed by the named label is the target of the jump. Therefore all jump targets can be known statically.
```
grosu: typo: "can used"

Design.md
bmmoore: I assume that calling `@deposit` in an account without code doesn't require passing any `gaslimit`. Do we force a `@deposit` function in a contract to do nothing (besides receiving balance), and allow it to be called with zero gas?

Design.md
grosu: I would add a separate bullet for `revert`.  What does it mean to revert the contract?  Is this also what throwing an exception does in the EVM?  How about the log, is it also restored?  How about the gas?  How is it implemented?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* The `ret` and `revert` instructions take an arbitrary number of register operands. `revert` reverts the contract, but `ret` will return to the instruction position on the top of the local call stack if one exists, restoring the values of local registers and copying the return values into the return value operands of the call instruction. If the local call stack is empty, it returns from the contract.
+* IELE has a `call .. at` and a `staticcall .. at` instruction used for calls to public functions of other contracts (the target account address is a register operand of the instruction, as is the amount of gas to be spent during the call, and the value to send in case of `call .. at`). `call .. at` and `staticcall .. at` take an arbitrary number of arguments and return values in the form of register operands instead of memory ranges. The arguments are copied into local registers corresponding to the formal arguments of the function being called. When the contract returns, the returned values are copied into the return value registers of the call instruction.
```
grosu: `staticall` ?

grosu: Is `call .. at` resembling something in EVM or LLVM?  I would be tempted to name it `call .. of`.  But I am not a native speaker, so @dwightguth and @bmmoore can decide

grosu: The long parentheses is too long; make it a normal sentence instead?

grosu: "When the contract returns"  Which one?  There are two involved.

Also, not clear what's the difference between the two call instructions.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* If the name of the function being called does not correspond to a public function of the contract being called, an exception occurs.
```
grosu: What does "exception" mean?  Revert?  We should probably have a discussion somewhere "On Exceptions" or something, and say that we do not have exceptions, we simply revert, although we may use the terminology `exception` or `throw`.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* In all cases, if a mismatch occurs between the number of arguments or return values of a function and the matching `call` or `ret` instruction, an exception occurs.
```
grosu: can't this be detected statically?  If not, then we should explain why.  But it feels like something that should be detectable statically; maybe after we add type declarations

Design.md
grosu: Can we provide any links here?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
 
 Unlike EVM, which uses 32-byte unsigned words, IELE has arbitrary-precision signed words.
 
 * Because words are now explicitly signed, the SDIV, SMOD, SLT, and SGT instructions are removed, and DIV, MOD, LT, and GT are replaced with signed versions.
-* An EXPMOD instruction is added to perform exponentiation modulo a particular base, and the MODEXP precompiled contract is removed.
-* SIGNEXTEND is changed to convert an N-byte twos-complement representation of a signed number into its signed value.
-* a TWOS instruction is added to convert a signed integer into an N-byte twos-complement representation of the number.
-* Because integers are arbitrary-precision, the integer operand of BYTE now counts from the least-significant byte instead of the most-significant.
+* An `expmod` instruction is added to perform exponentiation modulo a particular base, and the MODEXP precompiled contract is removed.
+* `sext` (corresponding to EVM's SIGNEXTEND) is changed to convert an N-byte twos-complement representation of a signed number into its signed value.
```
grosu: The name `sext` does not look good, but I do not have any better idea.  The EVM one looked better, more informative. 
 What is `N`?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* a `twos` instruction is added to convert a signed integer into an N-byte twos-complement representation of the number.
```
grosu: `twos` is not intuitive either.  Does EVM has a corresponding operation?

Since we'll write specifications at the IELE level, we want people to be able to read the programs naturally, without having to check the syntax descriptions or the semantics.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* Because integers are arbitrary-precision, the integer operand of `byte` (corresponding to EVM's BYTE) now counts from the least-significant byte instead of the most-significant.
```
grosu: "arbitrary precision"?  Maybe "unbounded" instead?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
 
-## Contract Creation
+## Instructions and Precompiled Contracts
+
+Unlike EVM, where each instruction is just an opcode operating in the values found in the top of the stack, IELE instructions follow a syntax similar to LLVM: each instruction is an opcode followed by some arguments in the form of registers. If the instruction produces a result, it is formed as an assignment where the LHS contains the register(s) to hold the result(s) and the RHS contains the opcode and the arguments. Other changes with respect to EVM include:
```
grosu: I would replace the two `in` on the first line with `on` and remove the `the` from `the top of`.  Let me formal:
` rule (in => on) ... (in the => on)`

Design.md
grosu: Then why don't we use this in the ERC20 implementation?  We should at least add a comment in the `erc20.iele`, @dwightguth

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
-* Transactions that create a contract now explicitly provide the binary data of the contract instead of the binary data of the initialization function. Contracts created in this way are not subject to the restriction that dynamic contract data be inserted, and still invoke the function at function label 0 to initialize the contract. If the contract is malformed, the transaction fails.
+## Contract Creation
 
-## Code Size
+Unlike EVM, both for ease of verification and due to security concerns, we disallow creation of dynamic code for dynamically created contracts, and instead restrict dynamically-created-contracts to code that consists of some previously validated contract.
 
-The maximum length of a IELE contract is expanded to 65536 bytes.
+* The `create` instruction is modified from its EVM counterpart. Instead of taking a memory start and a memory width containing initialization code, the `create` instruction takes a contract name, and an arbitrary number of register arguments to be passed to the `@init` function of the contract to be created. If the contract name is not an externally declared contract name, the contract is malformed.
```
grosu: This raises the issue of whether we want to have some static checker for malformedness.  I would put that at the end in the to-do list.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* Since each externally declared contract should have been defined in the same file and above the contract that externally declares it, the exact code that will be deployed by a `create` instruction is statically known and available.
+* The `create` instruction copies the corresponding contract code into the code storage of the account that contains the created contract. Then the `@init` function is invoked in the context of the created account to initialize the contract. It is passed the values in the registers as arguments in the same fashion as a `call .. at` or `staticcall .. at` instruction. If the `@init` function of a contract returns a value, the contract is malformed.
```
grosu: What if the created contract also wants to have external contracts to further create them?  Shouldn't we pass those external contracts it declares also with it?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* A separate instruction, `copycreate`, is introduced for the case of copying an entire contract, or duplicating your own contract. The `copycreate` instruction behaves like `create`, except that instead of taking an contract name, it takes a register operand containing an account address, and deploys the entire contract deployed to said account to the newly created account.
```
grosu: typo: "an contract"

grosu: Does EVM also have a similar copycreate command?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* Both instructions, similarly to EVM, return the address of the created contract on success and zero on failure.
```
grosu: I would emphasize that this is different from throwing, or reverting.  Also, why would this operation fail?  Not enough gas?  What happens if the account you want to deploy already has a contract in it?  Or belongs to somebody else?

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* Because we no longer actually need to address portions of code by program counter in order to create contracts, we remove the EVM's CODECOPY and EXTCODECOPY instructions. Because these are also the only remaining instructions which require the code to be addressed by a particular byte offset, we also remove the EVM's PC instruction.
+* Transactions that create a contract now explicitly provide the binary data of the contract instead of the binary data of the initialization function. Contracts created in this way are not subject to the restriction that dynamic contract data be inserted, and still invoke the `@init` function to initialize the contract. If the contract is malformed, the transaction fails.
```
grosu: Again, this raises the question: can we check malformedness statically?

Design.md
theo25: Not sure about that. The access for an initialized register in the semantics right now will return whatever the default value for K array elements is, and I don't know exactly what this is. Maybe @dwightguth can comment here.

Design.md
```diff
@@ -32,71 +32,73 @@ Here are the IELE design decisions described from the perspective of differences
...
+* An external contract declaration simply declares the name of another contract. The contract under definition can only create accounts with deployed code being a copy of the contract itself or one of the contracts that have been declared as external (see section on contract creation for details). Each externally declared contract should have been defined in the same file and above the contract that externally declares it.
+* A function definition includes the function signature, the function body and whether or not the function is public. A function signature includes a function name (which is also represented as a global register, as it is visible across the contract) and names of formal arguments (which are represented as local registers, as they are only visible within the function body).
+* A public function can be called by other accounts, while a non-public one can only be called by other functions within the same contract.
+* A special function named `@init` is automatically added to the contract (and can be overridden by the programmer), and will be called when an account is created with this contract.
```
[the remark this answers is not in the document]
theo25: It seems from the semantics that we actually do require a public init function to be defined. I will amend the text.

Design.md
theo25: The payment is independent of the code in the body of the function. If you see the syntax for account calls they look like: r1, r2 = call @FunName at Address ( Args ) send Value, gaslimit Gas. When doing an account call the amount in Value is transfered to the account as part of the message even if the called function does nothing. This is similar to EVM if I am not mistaken.
@bmmoore I think your understanding is correct.

Design.md
theo25: Fixed! (replaced by "mechanism")

Design.md
theo25: Actually, by reading the current semantics it is not clear to me what happens to global registers across account calls. But if they are not preserved, then what's the point of even having them?
Any comments @dwightguth ?

Design.md
theo25: I will fix that, but to be clear: our current design allows for a DAG (not necessarily a tree) of contracts with exactly one contract (the last one) not having any outgoing edges (meaning it's not declared in any other contract). That means that there is not a well defined notion of nesting. The only requirement is that the contracts are listed in their topological order. However it is true that the current binary encoding allows only trees.

Design.md
theo25: Actually both revert and return return values the same way, I will fix the text to reflect that.

Design.md
theo25: I am not sure we should do that. This would look to me too defensive, i.e. we ourselves suggesting solutions to problems before others raise them. If anything we should strengthen the argument that these are unsafe and should be removed at all costs.

Design.md
theo25: What do you mean by "we have still opcodes"? We have a notion of opcode in the binary encoding but that may change. As far as a programmer is concerned, these now behave as builtin functions that can be invoked from any contract with a local call. Why use a more generic language? This will create confusion in my opinion.

Design.md
theo25: This is a good question. Currently they can only be called as an account call, and I think that is the correct and intended way to call them. However such a call would look kind of bad in textual IELE because it would require to load the constant 1 as the address of the precompiled account:

```
%addr = 1
%val = ...
%gas = ...
%res  = call @iele.sha256 at %addr ( %len, %data ) send %val, gaslimit %gas
```
Maybe @dwightguth can comment on that too.

Design.md
theo25: I have mentioned the same in the first bullet of Program Structure section, but you make a fair point and I will repeat this here, as it will make it clearer to follow.

Design.md
theo25: Fixed!

Design.md
theo25: Throwing an exception and reverting are not exactly the same. The similarity is that when either an exception is raised or a revert is encountered while executing code in a function, the local call stack (the stack of local functions running right now) gets unwound and the control transfers out of the contract either to a calling contract deployed to another account or to the IELE VM if there is no such calling contract. Also, the active accounts and the log are restored to their state before the account call (in case one exists). The difference is that `revert` allows the values that it takes as arguments to be passed to the calling contract (in case one exists), while an exception does not return any values to the calling contract. Another difference is that `revert` is an instruction explicitly put in the program by the user, while exceptions correspond to runtime errors (e.g. division by zero).

Is this clear? Do you want me to explain all this in the doc? Also, I am not sure which parts of this are coming from EVM and which are new design. Maybe @dwightguth and @bmmoore e can help.

Design.md
theo25: Yes there is a STATICCALL opcode in EVM. `call .. at` refers to the account call instruction. I used `at` because the following operand is a register holding an account address, but I would also defer to @dwightguth and @bmmoore  for what works best linguistically.

theo25: Fixed! I didn't explain the difference because they both come from evm. A static account call, just like in EVM, should not try to change any state in the network (e.g. log, modify account storage, create/destruct accounts). If does attempt to do that an exception is raised. I can mention this if you want it in the text.

Design.md
theo25: We do have exceptions, see my reply above.

Design.md
theo25: It can be in case of local calls, but not always in case of account calls because the account address is not statically known and two accounts may have contracts deployed with public functions having the same name and different number of arguments. Simply knowing the function name statically does not help us in this case to detect this error.

Design.md
theo25: I used the same name as the LLVM instruction for signed integer extension. We can change it if everybody agrees. `N` is one of the arguments of the instruction. I will make this clear in the text.

Design.md
theo25: No, this one is new to IELE. I didn't pick the name and I don't have any suggestion (or strong feelings) on how to rename it.

Design.md
theo25: Fixed!

Design.md
theo25: Fixed!

Design.md
theo25: Yes, I think @dwightguth has mentioned the same, that he wants to have one checker for well-formed IELE programs in the future. Currently the semantics assumes only well-formed programs.

Design.md
theo25: Yes,  we actually do. I will mention that in the text.

Design.md
theo25: Fixed!

theo25: No. This was a design decision we made for the case of a contract wanting to deploy a copy of itself.

Design.md
theo25: `create` and `copycreate` always create a new account, hence it cannot have a contract already deployed in it. Various things can make a contract creation fail, e.g. the @init function of the contract to be created raises an exception (or not enough gas as you mentioned).

Design.md
theo25: As said above, yes we should be able to check malformedness statically.

PR #30, opened 2017-12-01
Fix KRYPTO.ml for newest secp256k1 version 0.3.2

bmmoore:
Key recovery now returns an option. This makes `KRYPTO.ml` compile again just by assuming recovery succeeds.

### DIFF
--- a/KRYPTO.ml
+++ b/KRYPTO.ml
@@ -41,7 +41,7 @@ let hook_ecdsaRecover c lbl sort config ff = match c with
     let messageBuffer = Bigarray.Array1.of_array Bigarray.char Bigarray.c_layout messageArray in
     let publicKey = match Secp256k1.RecoverableSign.recover context signature messageBuffer with
         Some k -> k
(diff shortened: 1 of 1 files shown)


KRYPTO.ml
```diff
@@ -39,7 +39,7 @@ let hook_ecdsaRecover c lbl sort config ff = match c with
     if String.length hash <> 32 then [String ""] else
     let messageArray = Array.init 32 (fun idx -> hash.[idx]) in
     let messageBuffer = Bigarray.Array1.of_array Bigarray.char Bigarray.c_layout messageArray in
-    let publicKey = Secp256k1.RecoverableSign.recover context signature messageBuffer in
+    let publicKey = match Secp256k1.RecoverableSign.recover context signature messageBuffer with Some k -> k in
```
dwightguth: if this returns None you need to return [String ""] from the entire function

bmmoore: Ok, looks like the existing handler will do that if l throw a `Failure` exception on `None`.

PR #32, opened 2017-12-02
Installation instructions

dwightguth:
Tested on a fresh Ubuntu 16.04.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,19 @@
+Installing IELE
+===============
+
+In a nutshell, to install IELE on an Ubuntu 16.04 machine, run:
+
+```
+
+sudo apt-get update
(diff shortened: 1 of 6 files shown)


INSTALL.md
```diff
@@ -0,0 +1,19 @@
...
+===============
+
+In a nutshell, to install IELE on an Ubuntu 16.04 machine, run:
+
+```
+
+sudo apt-get update
+sudo apt-get install make gcc maven openjdk-8-jdk flex opam pkg-config libmpfr-dev autoconf libtool
+git submodule update --init # Initialize submodules
+curl -sSL https://get.haskellstack.org/ | sh # Install stack
```
bmmoore: Was curl installed by default? It isn't in the LXC image of ubuntu 16.04

PR #35, opened 2017-12-06
Avoid lua-5.2-only function in tangle.lua

bmmoore:
It turns out table.unpack was added in 5.2 and the
pandoc distributed in Ubuntu 14.04 was built with 5.1

### DIFF
--- a/tangle.lua
+++ b/tangle.lua
@@ -43,7 +43,8 @@ function Doc(body, metadata, variables)
   local targets = split_classes(metadata.code or '')
   local block = 1
   local function replace()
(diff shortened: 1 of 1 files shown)


dwightguth:
16.04 actually... 14.04 may be even older.

bmmoore:
Ah, I assumed it was older because I tested the semantics in 16.04, but I was only testing your install instructions before we switched the tangle script.

bmmoore:
The earliest pandoc release whose release notes mentions lua writers also used Lua 5.1 so there shouldn't be any more compatibility problems in this code.

PR #36, opened 2017-12-07
Assembler

bmmoore:
Update assembler so globals are just named constants, so create and copycreate return separate results for status and new address, and to correctly assemble external contract declarations.


dwightguth:
What about the change to revert?

PR #37, opened 2017-12-07
Fixes for negative numbers, revert, create, copy create

dwightguth:
Contains several fixes:

* fix IELE semantics for negative operands
* implement refactoring of revert and exit statuses of calls.
    * revert takes one operand and returns it in the status register
    * exit statuses are:
        * 0: success
        * 1: function does not exist
        * 2: function has wrong signature
        * 3: function does not exist on empty account
        * 4: execution of instructions led to failure
        * 5: out of gas
        * 6: deploying to an account that already exists
        * 7: insufficient balance to transfer
        * 8: negative balance or gas limit or call depth exceeded
* remove global registers from semantics (they are now syntactic sugar for immediates)
* add tests for create and copycreate and fix bugs in implementation
* add documentation to configuration
* clean up and simplify test harness


iele.md
```diff
@@ -1569,7 +1580,7 @@ Precompiled Contract
     rule <k> #exec ECPAIRING => #ecpairing(.List, .List, DATA) ... </k>
          <callData> DATA </callData>
       requires #sizeRegs(DATA) %Int 6 ==Int 0
-    rule <k> ECPAIRING => #exception ... </k>
+    rule <k> ECPAIRING => #exception 2 ... </k>
```
bmmoore: Are you planning to make this have a fixed arity?

PR #39, opened 2017-12-11
Support global definitions, immediates, output options

bmmoore:
Now allows immediate arguments to any instruction, which
is compiled by loading the immediates into temporary registers.

Allow global definitions of named constants, which expand
into immediate values.

Add options to reprint the input, either immediately after parsing
or just before assembling.

### DIFF
--- a/compiler/README.md
+++ b/compiler/README.md
@@ -1,7 +1,8 @@
 # compiler
 To build: `stack build`.
-all dependencies are managed by stack
+all dependencies are managed by stack.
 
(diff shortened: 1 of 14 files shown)


compiler/README.md
```diff
@@ -1,6 +1,7 @@
 # compiler
 To build: `stack build`.
+all dependencies are managed by stack
 
-To run: `stack exec compiler-exe`.
+To run: `stack exec iele-assmble`.
```
dwightguth: s/assmble/assemble

PR #40, opened 2017-12-12
Gasink

dwightguth:
Should be the last pull request aside from any polishing to be done to documentation or extremely urgent changes. Basically just contains the new gas model.

### DIFF
--- a/iele-gas.md
+++ b/iele-gas.md
@@ -739,7 +739,7 @@ Note: These are all functions as the operator `#compute` has already loaded all 
     rule Ckara(L1, L2) => L1 *Int #overApproxKara(L2) /Int L2
       requires L1 >=Int L2
 
-    rule Ckara(L1, L2) => L2 *Int #overApproxKara(L1) /Int L1
+    rule Ckara(L1, L2) => L2 *Int #overApproxKara(L1) /Int L1 [owise]
 
     rule Cdiv(SCHED, L1, L2) =>
         Gdivkara < SCHED > *Int Ckara(L1 -Int L2 +Int 1, L2) +Int
--- a/iele-gas.md
+++ b/iele-gas.md
@@ -715,29 +715,29 @@ Note: These are all functions as the operator `#compute` has already loaded all 
                  | Cexpmod  ( Schedule , Int , Int , Int ) [function]
  // -----------------------------------------------------------------
     rule Cmul(SCHED, L1, L2) => Cmul(SCHED, L2, L1)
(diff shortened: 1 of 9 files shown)


dwightguth:
Not quite done yet. Still needs mul, div, exp, and expmod formulas

iele-gas.md
```diff
@@ -720,47 +720,57 @@ Note: These are all functions as the operator `#compute` has already loaded all
...
+      requires L1 >=Int L2
 
-    rule Cexpmod(SCHED, LB, LEX, LM) =>
-        Cdiv(SCHED, LB, LM) +Int
-        Gexpmodmul < SCHED > *Int LEX *Int Cmul(SCHED, LM, LM) +Int
-        Cdiv(SCHED, 2 *Int LM, LM) +Int
-        Gexpmodwordm < SCHED > *Int LM +Int
-        Gexpmodworde < SCHED > *Int LEX +Int
-        Gexpmod < SCHED >
+    rule Ckara(L1, L2) => L2 *Int #overApproxKara(L1) /Int L1
```
bmmoore: should this have `[owise]` or a `requires` clause?

dwightguth: what's wrong with owise?

bmmoore: It would be fine, but this rule doesn't have `owise`, does it?

iele-gas.md
```diff
@@ -0,0 +1,960 @@
...
+* `a1 = 2 * 32 - 32/2 = 48`, and `b1 = 1024/2 + a1 - 1024 / 8 = 432`
+* next, the approximation must also be continuous, so
+* `a0 = 32^2 - 32^2/4 - 48*32 = -768`, and `b0 = 1024^2/4+ 48*1024 -768 - 1024^2/16 - 432*1024 = - 197376`
+
+```{.k .uiuck .rvk}
+    syntax Int ::= #overApproxKara ( Int )                 [function]
+ // -----------------------------------------------------------------
+    rule #overApproxKara(N) => #if N <=Int   32 #then N *Int N
+                         #else #if N <=Int 1024 #then N *Int N /Int  4 +Int  48 *Int N -Int    768
+                         #else                        N *Int N /Int 16 +Int 432 *Int N -Int 197376
```
bmmoore: We should make the final case actual behave asymptotically as `x^log_2(3)` or at least a closer overapproximation like `x^1.6`, rather than as `x^2` like this term. What is the upper bound to validity of this last polynomial?

PR #50, opened 2017-12-14
Change test instructions to use iele-test target

bmmoore:
Now suggests testing iele tests with iele-test target
and describes how to run individual tests, rather than
only suggesting `make test` or explicit shell loops over
the testcases.


dwightguth:
@bmmoore you got the filename of simpleOpenAuction wrong

PR #52, opened 2017-12-14
INSTALL.md: windows and non-ubuntu instructions

dwightguth:


INSTALL.md
```diff
@@ -20,3 +20,7 @@ make # Build project
 ```
 
 To test against a fragment of the ethereum test suite, run `make test`
+
+Other Linux distros can install by installing the list of dependencies in the `apt-get install` command above and then following the rest of the instructions.
+
+To install on Windows, first install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10) and then proceed with the installation for Ubuntu 16.04. Note that the installation will take significantly longer due to existing problems in the performance of disk-intensive processes in WSL, and you should expect it to take quite some time. However, once installed, it should run roughly the same speed as on native Linux.
```
bmmoore: "Note that the installation will take significantly longer ... and you should expect it to take quite some time." is fairly redundant. I would change the last clause to "and you should be prepared for it to take at least half an hour".

bmmoore:
This has been tested. It works, it just takes a long time to install.

PR #55, opened 2017-12-19
Changes to make kast work

traiansf:
This should be enough to make kast work.

apaprently `kast` is looking for a `<MODULE>-PROGRAM-PARSING` and if it does not find one, it generates one including `ID-PROGRAM-PARSING` syntax which contains some token rules with higher precedence than ours.  That is why the `IeleName`s ended up being parsed as `Id`s and `kast` reported an error.


traiansf:
This should be merged only after https://github.com/runtimeverification/rv-match/pull/641

dwightguth:
Merge this tomorrow

PR #59, opened 2018-01-25
fix bugs in install process

dwightguth:
Fixes an issue reported by a user on the public iele channel. @theo25 please review.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,7 +6,7 @@ In a nutshell, to install IELE on an Ubuntu 16.04 machine, run:
 ```
 
 sudo apt-get update
-sudo apt-get install make gcc maven curl openjdk-8-jdk flex opam pkg-config libmpfr-dev autoconf libtool pandoc libssl-dev build-essential libffi-dev libpython-dev python-setuptools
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -6,7 +6,7 @@ In a nutshell, to install IELE on an Ubuntu 16.04 machine, run:
 ```
 
 sudo apt-get update
-sudo apt-get install make gcc maven curl openjdk-8-jdk flex opam pkg-config libmpfr-dev autoconf libtool pandoc libssl-dev build-essential libffi-dev libpython-dev python-setuptools
+sudo apt-get install make gcc maven curl openjdk-8-jdk flex opam pkg-config libmpfr-dev autoconf libtool pandoc libssl-dev build-essential libffi-dev libpython-dev python-setuptools libssl-dev
 git submodule update --init # Initialize submodules
```
darko-marinov: I undetstand this was already approved/merged, but it seems you already had libssl-dev earlier in the list: does the order matter? Should the earlier occurrence be removed? It looks a bit sloppy if we repeat this unnecessarily.

PR #61, opened 2018-01-30
update to latest version of k

dwightguth:


krypto.md
```diff
@@ -5,7 +5,7 @@ Here we implement the various cryptographic primitives needed for IELE.
 
 ```{.k .uiuck .rvk}
 module KRYPTO
-    imports STRING-SYNTAX
+    imports LIST
```
bmmoore: Why did `STRING-SYNTAX` change to `LIST` here? Was `STRING-SYNTAX` just very wrong?

PR #62, opened 2018-01-31
Syntax highlighting of IELE for VIM.

shd101wyy:
screen shot 2018-01-31 at 3 35 16 pm

To install, copy (or link) "iele.vim" file into the ~/.vim/syntax directory and add the following to your ~/.vimrc file

au BufRead,BufNewFile *.iele set filetype=iele au! Syntax iele source iele.vim syn on

---

Please try it first. Thank you!


dwightguth:
it seems to me that `@sum` should be the same color as `@iele.invalid`, shouldn't it? they're both function names.

shd101wyy:
@dwightguth How about this now?

screen shot 2018-01-31 at 4 10 27 pm

dwightguth:
I think it looked better with different colors for functions and registers, and it was definitely better when the @ symbol was colored the same color as the name. Also, you shouldn't reuse the same colors between register names and contract names. My personal preference also was that the green looked better for the function names, but I'm open to be persuaded of alternatives.

shd101wyy:
screen shot 2018-01-31 at 5 19 52 pm

> I think it looked better with different colors for functions and registers, and it was definitely better when the @ symbol was colored the same color as the name

Done

> Also, you shouldn't reuse the same colors between register names and contract names.

This is related to the vim theme. Different themes will use different colors. I configured contract name to use color of `Type`, and register name to use color of `Identifier`. Seems that my vim theme uses the same color for both `Type` and `Identifier`.  

> My personal preference also was that the green looked better for the function names, but I'm open to be persuaded of alternatives.

Done

shd101wyy:
I changed the style again and fixed the highlighting of variable like `@"name"`.

screen shot 2018-08-10 at 2 53 46 pm

SebastienGllmt:
Thought: This seems like something that should be done at the K-Framework level and then annotated in the IELE semantics through tags (similar to how latex is handled). It's harder to do but should be more maintainable because this version introduces code duplication and therefore will almost certainly cause version mismatch issues.

PR #67, opened 2018-02-02
Node pb code

bmmoore:
This adds server and client code for connecting IELE to a node.
Everything compiles, including adjusting the makefile so the `ieleVM.ml` could run the new server code. (The makefile changes are all in the first commit, the last commit only touches `ieleVM.ml`, to confirm the changes to the `iele-vm` target are good, by hacking up `ieleVM.ml` with simple and incorrect changes to fix compilation errors).

@dwightguth This probably needs to be rebased or cherry-picked onto your work instead of merged as a pull request. You probably don't want f025482, and may not want to include 60f7561 in your node branch.


bmmoore:
I have updated this against the rebased version of the `node` branch, making a different PR.

PR #68, opened 2018-02-08
Mantis client and server networking code

bmmoore:
Communication isn't tested at all, but the server listens when run.


bmmoore:
@dwightguth I cleaned up the comments, and also extracted the caching code as we were talking about. How does this look?

PR #72, opened 2018-02-14
Move iele-semantics-plugin and iele-vm directories to git subtree to share with EVM

dwightguth:


plugin/vm/VM.ml
```diff
@@ -83,54 +70,104 @@ let k_to_mod_acct (acct: k) : modified_account = match acct with
...
+  [KApply0(parse_klabel "EIP158_EVM")]
+
+let get_output_data k_rets config = match config with
+| Iele_config ->
+  begin
+    match k_rets with [List(SortList,Lbl_List_,k_rets)] ->
+    let z_rets = List.map k_to_z k_rets in
+    let rets = List.map World.of_z z_rets in
+    pack_output rets
+    | _ -> failwith "Invalid valuer where list was expected"
```
bmmoore: "Invalid valuer"

plugin/vm/VM.ml
```diff
@@ -83,54 +70,104 @@ let k_to_mod_acct (acct: k) : modified_account = match acct with
...
+    else
+      Bytes.to_string code,args,""
+  end
 
 let pack_output rets =
   let l = List.map (fun ret -> Rlp.RlpData (Rope.of_string (Bytes.to_string ret))) rets in
   let rlp = Rlp.RlpList l in
   Bytes.of_string (Rope.to_string (Rlp.encode rlp))
 
+let get_schedule blocknum config = match config with
```
bmmoore: It looks like you are trying to handle the differences between EVM and IELE by writing code that takes a flag, and being careful that the unused branch will at least compile against the `constants.ml` of the other definition.

I suppose it's okay for now, but we can't count on that in the long term. Instead, this code could be wrapped into a functor to be instantiated with definition-specific implementations when it's used from the IELE and EVM semantics.

dwightguth:
Closing in favor of #73

PR #77, opened 2018-02-22
Downgrade version of secp256k1 to install

theo25:
The secp256k1 opam package seems to still be in flux with respect to its public API.

We are fixing this to a specific version so that if it changes again, it will not break
new installs.


dwightguth:
Jenkins: test this please

PR #82, opened 2018-02-28
README.md update to point to iele-syntax.nd and Design.md

theo25:
Updated README.md to point to iele-syntax.md as the starting point forr familiarizing with IELE.


theo25:
Feel free to modify the language if you want.

PR #89, opened 2018-03-09
python scripts: Change shebang to call `python2` instead of `python`

nishantjr:
This allows running of IELE code on machines with both python2 and python3
installed.

### DIFF
--- a/assemble-iele-test
+++ b/assemble-iele-test
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 import subprocess
 import sys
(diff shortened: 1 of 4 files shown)


dwightguth:
This is not portable as not all os that install python 2 install a binary named python2. What OS are you trying to support?

nishantjr:
hmm, I'm on archlinux. Which OS's dont do install the `python2` binary, do you know?

dwightguth:
Mac OS for one. If I have to choose between supporting Mac OS and Arch Linux, I'm going to choose Mac OS, so if you want to support Arch Linux you need to come up with a different solution.

nishantjr:
Sure, I'll work on getting the scripts supporting both python 2 and python 3 at some point.

PR #98, opened 2018-03-20
Add iele-test-client binary to support Solidity-to-IELE harness

theo25:


theo25:
Jenkins: test this please

PR #113, opened 2018-04-11
Binary logarithm instruction

dwightguth:
Adds a log2 instruction which computes the binary logarithm of an integer.


iele-binary.md
```diff
@@ -49,6 +49,7 @@ contains enough information to determine and decode the rest of the operation.
                    | MLOAD ()
                    | MSTORE ()
                    | SLOAD ()
+                   | LOGARITHM ()
```
bmmoore: Maybe this name should also mention the base.

tests/iele/unit/logarithm.iele.json
```diff
@@ -0,0 +1,123 @@
...
+                    {
+                        "nonce": "0x00", 
+                        "function": "test_logarithm", 
+                        "gasLimit": "0x100000", 
+                        "value": "0x00", 
+                        "to": "0x1000000000000000000000000000000000000000", 
+                        "arguments": ["0x04"], 
+                        "contractCode": "",
+                        "gasPrice": "0x01",
+                        "from": "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b"
```
bmmoore: There should be another test on input larger than a byte, maybe something larger than 64 bits.

dwightguth:
@bmmoore please re-review

bmmoore:
Test is good now. I asked for a little more explanation of the cost in `gas.md` - say it's rounded down so it can be computed by finding the position of the highest set bit.

PR #114, opened 2018-04-11
allow escaped iele names as names of contracts in parser

dwightguth:
This is needed because we need to qualify the names of contracts in compiled solidity code by their filename, and filenames can contain arbitrary data.


well-formedness.md
```diff
@@ -320,6 +320,22 @@ All identifiers beginning with "iele." are reserved by the language and cannot b
...
+
+```k
+    syntax String ::= unescape(String) [function]
+                    | unescape(String, Int, StringBuffer) [function, klabel(unescapeAux)]
+ // -------------------------------------------------------------------------------------
+    rule unescape(S) => unescape(S, 1, .StringBuffer)
+    rule unescape(S, IDX, SB) => unescape(S, IDX +Int 1, SB +String substrString(S, IDX, IDX +Int 1))
+      requires IDX <Int lengthString(S) -Int 1 andBool substrString(S, IDX, IDX +Int 1) =/=K "\\"
+    rule unescape(S, IDX, SB) => unescape(S, IDX +Int 3, SB +String chrChar(String2Base(substrString(S, IDX +Int 1, IDX +Int 3), 16)))
+      requires IDX <Int lengthString(S) -Int 1 andBool substrString(S, IDX, IDX +Int 1) ==K "\\"
```
theo25: I think the first part of `requires` here should be `IDX <Int lengthString(S) -Int 3`

theo25: Never mind. It should be fine as is since the strings are guaranteed to be well-formed because they have been checked with the parser.

dwightguth: We decided offline to leave it as is. The code is guaranteed to work correctly for well formed strings, which we are guaranteed to be operating on because the strings have already been checked by the parser

PR #118, opened 2018-04-20
fix bug in bswap when N > bytewidth(W)

dwightguth:


dwightguth:
We all missed it, don't worry.

PR #123, opened 2018-04-30
Function pointers in IELE

dwightguth:
Per our previous discussion, this PR adds function pointers to IELE by allowing global function names anywhere an operand can be used in the assembler, and adding semantics for dynamic call targets computed from an integer function offset.

### DIFF
--- a/compiler/src/IeleAssembler.hs
+++ b/compiler/src/IeleAssembler.hs
@@ -112,8 +112,11 @@ asm_iele_opcode_li opLi = case opLi of
 asm_iele_opcode_call :: IeleOpcodeCall Word16 Word16 -> Put
 asm_iele_opcode_call opCall = case opCall of
   CALL call nargs nreturn -> putWord8 0xf2 >> putWord16be call >> putArgs16 nargs >> putRets16 nreturn
+  CALLDYN nargs nreturn -> putWord8 0xf3 >> putArgs16 nargs >> putRets16 nreturn
   STATICCALL call nargs nreturn -> putWord8 0xf5 >> putWord16be call >> putArgs16 nargs >> putRets16 nreturn
+  STATICCALLDYN nargs nreturn -> putWord8 0xf9 >> putArgs16 nargs >> putRets16 nreturn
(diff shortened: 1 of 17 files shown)


nishantjr:
@dwight possible to add a test for storing a function pointer and retrieving it from memory?
Something along the lines of

```
define @test() {
   %func = @f1
   store %func, 1234
   %loadedFunc = load 1234
   call %loadedFunc()
}

define @f1() {
  ret 42
}
```

dwightguth:
@nishantjr added

compiler/src/IeleAssembler.hs
```diff
@@ -112,8 +112,11 @@ asm_iele_opcode_li opLi = case opLi of
 asm_iele_opcode_call :: IeleOpcodeCall Word16 Word16 -> Put
 asm_iele_opcode_call opCall = case opCall of
   CALL call nargs nreturn -> putWord8 0xf2 >> putWord16be call >> putArgs16 nargs >> putRets16 nreturn
+  CALLDYN nargs nreturn -> putWord8 0xf3 >> putArgs16 nargs >> putRets16 nreturn
   STATICCALL call nargs nreturn -> putWord8 0xf5 >> putWord16be call >> putArgs16 nargs >> putRets16 nreturn
+  STATICCALLDYN nargs nreturn -> putWord8 0xf9 >> putArgs16 nargs >> putRets16 nreturn
   LOCALCALL call nargs nreturn -> putWord8 0xf8 >> putWord16be call >> putArgs16 nargs >> putRets16 nreturn
+  LOCALCALLDYN nargs nreturn -> putWord8 0xf4 >> putArgs16 nargs >> putRets16 nreturn
```
bmmoore: Why not have f4 for staticcalldyn, nearer to f5 for staticcall and f9 for localcalldyn, near to f8 for localcall

compiler/src/IeleDesugar.hs
```diff
@@ -44,8 +44,8 @@ expandImmediates inst =
   in reverse loads++[inst']
 
 desugarFundef :: Map GlobalName Integer
-              -> FunctionDefinitionP
```
bmmoore: Could the type synonym be changed so it is still usable here?

PR #124, opened 2018-05-02
Get function offset instruction

dwightguth:
Adds an instruction that gets the function pointer of an external function from its name and an address.

### DIFF
--- a/compiler/src/IeleInstructions.hs
+++ b/compiler/src/IeleInstructions.hs
@@ -188,6 +188,7 @@ data IeleOpcodeCall contractId funId =
  | STATICCALLDYN (Args Word16) (Rets Word16)
  | LOCALCALL funId (Args Word16) (Rets Word16)
  | LOCALCALLDYN (Args Word16) (Rets Word16)
+ | CALLADDRESS funId
 
  | CREATE contractId (Args Word16) -- contract Id
  | COPYCREATE (Args Word16)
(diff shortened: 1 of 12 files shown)


tests/iele/unit/fptr.iele
```diff
@@ -27,6 +27,8 @@ contract test {
     revert -1
   }
 
+  @unused = 13
+
```
theo25: Why was that added?

dwightguth: Testing that we still compute the right offset for test2

compiler/src/IelePrint.hs
```diff
@@ -162,6 +162,13 @@ prettyIeleInst (CallOp (STATICCALLDYN _ _) results allArgs) = case results of
              <> char '(' <> commaList args <> char ')'
              <+> text "gaslimit" <+> gas
          _ -> error "external staticcall instruction must encode at least target and gaslimit arguments"
+prettyIeleInst (CallOp (CALLADDRESS name) results allArgs) = case results of
+  [] -> error "external call instruction must have at least one result"
+  _ -> case allArgs of
+         (acct:[]) ->
+           prettyResults results <+> text "calladdress" <+> prettyName name
+             <+> text "at" <+> acct
+         _ -> error "external staticcall instruction must encode at least target argument"
```
bmmoore: This message says "staticcall". This should also expect exactly one result, right?

compiler/src/IeleInstructions.hs
```diff
@@ -205,6 +206,7 @@ retypeOpcodeCall contract fun i = case i of
     STATICCALLDYN nargs nrets -> pure (STATICCALLDYN nargs nrets)
     LOCALCALL f nargs nrets -> (\f -> LOCALCALL f nargs nrets) <$> fun f
     LOCALCALLDYN nargs nrets -> pure (LOCALCALLDYN nargs nrets)
+    CALLADDRESS f -> (\f -> CALLADDRESS f) <$> fun f
```
bmmoore: This could be `CALLADDRESS <$> fun f`

PR #127, opened 2018-05-07
Metadata stamp for binary representation.

theo25:
This PR adds a new opcode in the binary format (0x6b) that signifies start of a metadata stamp. The metadata stamp is a binary string of arbitrary length and contents that is being dropped from the IELE disassembler at loading time. I didn't include this opcode in the assembler since there is no way to add a metadata stamp to a textual iele program.

### DIFF
--- a/iele-binary.md
+++ b/iele-binary.md
@@ -134,7 +134,6 @@ After interpreting the strings representing programs as a `WordStack`, it should
     rule #dasmContract( .WordStack, _) => #emptyCode
     rule #dasmContract( W1 : W2 : W3 : W4 : 99 : WS, NAME) => #dasmContract(99 : #take( W1 *Int 16777216  +Int W2 *Int 65536 +Int W3 *Int 256 +Int W4 -Int 1, WS), NAME, W1 : W2 : W3 : W4 : .WordStack)
     rule #dasmContract( 99 : NBITS : WS, NAME, W1 : W2 : W3 : W4 : .WordStack ) => #dasmContract(WS, NBITS, 0 |-> init, NAME, .TopLevelDefinitions, 1, #sizeWordStack(WS) +Int 6 , #unparseByteStack(W1 : W2 : W3 : W4 : 99 : NBITS : WS))
-    rule #dasmContract( 99 : NBITS : WS, NAME ) => #dasmContract(WS, NBITS, 0 |-> init, NAME, .TopLevelDefinitions, 1, #sizeWordStack(WS) +Int 2 , #unparseByteStack(99 : NBITS : WS))
     rule #dasmContract( 105 : W1 : W2 : WS, NBITS, FUNCS, NAME, DEFS, N, SIZE, BYTES ) => #dasmContract(#drop(W1 *Int 256 +Int W2, WS), NBITS, N |-> #parseToken("IeleName", #unparseByteStack(#take(W1 *Int 256 +Int W2, WS))) FUNCS, NAME, DEFS, N +Int 1, SIZE, BYTES )
     rule #dasmContract( 106 : W1 : W2 : WS, NBITS, FUNCS, NAME, DEFS, N, SIZE, BYTES ) => #dasmContract(#take(W1 *Int 256 +Int W2, WS), NAME +.+IeleName N) ++Contract #dasmContract(#drop(W1 *Int 256 +Int W2, WS), NBITS, FUNCS, NAME, external contract NAME +.+IeleName N DEFS, N +Int 1, SIZE, BYTES)
     rule #dasmContract( WS, NBITS, FUNCS, NAME, DEFS, N, SIZE, BYTES ) => contract NAME ! SIZE BYTES { DEFS ++TopLevelDefinitions #dasmFunctions(WS, NBITS, FUNCS, NAME) } .Contract [owise]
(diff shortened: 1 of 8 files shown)


bmmoore:
Should this be documented anywhere?
It looks like this just drops the rest of the file after the opcode is encountered. That seems like a rather inflexible way to allow for metadata. Why does it need to consume opcode space at all, rather than being part of a header or string table or something like that?

dwightguth:
@bmmoore Solidity assumes that the metadata is at the end of the contract. We could maybe break that assumption but it's hard for me to know what all tools out there depend on that assumption, and it's going to be easier if we keep it in the same place in the bytecode, because we won't have to change any other tools.

bmmoore:
Why does this need a new opcode?

theo25:
How would you suggest doing it without a new opcode? I don’t have something in mind

theo25:
To be clear: I really dislike the idea of metadata being equal citizens with strings

bmmoore:
I don't mind modifying the binary encoding if there's a good reason, but I don't want to consume an opcode. I think this belongs more as "object file format" than "instruction encoding", if we imagine the binary format made a strict distinction

theo25:
We agreed offline to add the size of the contract without the metadata as a 4 byte word in the header of the contract. The PR is updated to reflect this design.

theo25:
This is not ready for review yet since it' buggy, but could you elaborate on why it would be inefficient?

dwightguth:
I take it back, I was concerned about stack depth and paying a linear cost, but #take is implemented as tail recursive and we are paying a linear cost either way.

iele-binary.md
```diff
@@ -131,6 +131,7 @@ After interpreting the strings representing programs as a `WordStack`, it should
                       | #dasmContract ( WordStack , Int , Map, IeleName , TopLevelDefinitions, Int , Int , String ) [function, klabel(#dasmContractAux)]
  // ----------------------------------------------------------------------------------------------------------------------------------------------------
     rule #dasmContract( .WordStack, _) => #emptyCode
+    rule #dasmContract( W1 : W2 : W3 : W4 : 99 : WS, NAME) => #dasmContract(99 : #take( W1 *Int 16777216  +Int W2 *Int 65536 +Int W3 *Int 256 +Int W4, WS), NAME)
```
bmmoore: This `#take` isn't counting the `99` byte. The assembler included all bytes in the length, and I think it's best to count all the data that's meant to be part of the contract.

theo25: I am actually just fixing this :)

iele-binary.md
```diff
@@ -128,9 +128,12 @@ After interpreting the strings representing programs as a `WordStack`, it should
...
+                      | #dasmContract ( WordStack , IeleName , WordStack ) [function, klabel(#dasmContractAux1)]
+                      | #dasmContract ( WordStack , Int , Map, IeleName , TopLevelDefinitions, Int , Int , String ) [function, klabel(#dasmContractAux2)]
  // ----------------------------------------------------------------------------------------------------------------------------------------------------
     rule #dasmContract( .WordStack, _) => #emptyCode
+    rule #dasmContract( W1 : W2 : W3 : W4 : 99 : WS, NAME) => #dasmContract(99 : #take( W1 *Int 16777216  +Int W2 *Int 65536 +Int W3 *Int 256 +Int W4 -Int 1, WS), NAME, W1 : W2 : W3 : W4 : .WordStack)
+    rule #dasmContract( 99 : NBITS : WS, NAME, W1 : W2 : W3 : W4 : .WordStack ) => #dasmContract(WS, NBITS, 0 |-> init, NAME, .TopLevelDefinitions, 1, #sizeWordStack(WS) +Int 6 , #unparseByteStack(W1 : W2 : W3 : W4 : 99 : NBITS : WS))
     rule #dasmContract( 99 : NBITS : WS, NAME ) => #dasmContract(WS, NBITS, 0 |-> init, NAME, .TopLevelDefinitions, 1, #sizeWordStack(WS) +Int 2 , #unparseByteStack(99 : NBITS : WS))
```
dwightguth: pretty sure this rule needs to be removed as it conflicts with the rules added above.

theo25: hmm, yes you are right that it is in conflict with the other rules, I missed that. I kept this rule to not have to update the ethereum tests that have hardcoded code bytestrings. But I guess we should remove the rule since it's in conflict.

dwightguth: it shouldn't be that hard to modify the evm-to-iele translator to add the length

theo25: yes that's what I am doing

tests/evm-to-iele/iele.ml
```diff
@@ -192,6 +192,18 @@ let asm_iele ops =
...
+    Bytes.init 4 nth_byte_be
+  in
+  let bytes_buf = Buffer.create ((List.length ops) * 2) in
+  asm_iele_aux ops bytes_buf nregs;
+  let bytes_length = Buffer.length bytes_buf in
+  if bytes_length = 0 then
+    ""
+  else
+    let buf = Buffer.create (bytes_length + 4) in
+    Buffer.add_bytes buf (int_to_bytes bytes_length);
```
dwightguth: can you use the be_int_width function that already exists for this?

theo25: fixed!

PR #130, opened 2018-05-17
Binary contract validator

### LINKED ISSUE #121: Iele binary well-formedness
The well-formedness semantics checks that a disassembled iele contract is well-formed, but we also need to check that a iele contract can be disassembled successfully.

dwightguth:
Code that checks to make sure we can successfully disassemble the contract prior to actually trying to disassemble it.

This prevents crashes in cases where we previously had not given semantics, but also closes attack vectors where previously large immediate constants could have led to an improperly large amount of time spent trying to disassemble a maliciously constructed contract.

Fixes #121

Ready for review.

### DIFF
--- a/iele-binary.md
+++ b/iele-binary.md
@@ -42,6 +42,7 @@ contains enough information to determine and decode the rest of the operation.
                     | MSIZE ()
                     | SELFDESTRUCT ()
                     | LOG0 ()
+                    | REVERT()
 
     syntax BinOp ::= ISZERO ()
                    | NOT ()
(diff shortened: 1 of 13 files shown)


iele-binary.md
```diff
@@ -131,13 +131,29 @@ After interpreting the strings representing programs as a `WordStack`, it should
                       | #dasmContract ( WordStack , IeleName , WordStack ) [function, klabel(#dasmContractAux1)]
                       | #dasmContract ( WordStack , Int , Map, IeleName , TopLevelDefinitions, Int , Int , String ) [function, klabel(#dasmContractAux2)]
  // ----------------------------------------------------------------------------------------------------------------------------------------------------
-    rule #dasmContract( .WordStack, _) => #emptyCode
+    rule #dasmContract( .WordStack, NAME) => #emptyCode
```
theo25: nit: `NAME` is not used so it can be replaced with `_` as before

PR #136, opened 2018-06-20
Gas model revision

dwightguth:
Finally ready for review.


iele-gas.md
```diff
@@ -576,19 +591,18 @@ constant cost to perform the jump and store the return address.
       requires notBool IDX in_keys(LABELS)
 ```
 
-The cost to return from an intra-contract call is the cost to move the return values into the result registers plus the cost to
-jump to the return address.
+The cost to return is folded into the cost of the call instruction.
```
bmmoore: This comment looks like it is only correct when `Gnewmove << SCHED >>` is true.

iele-gas.md
```diff
@@ -601,10 +615,10 @@ The cost to call another contract is very similar to the cost in EVM:
 -   The gas stipend paid to the callee to execute its code.
 
 ```k
-    rule <k> #compute [ _, RETS::LValues = call _ at ACCTTO ( ARGS ) send VALUE , gaslimit GCAP, SCHED ] => Ccall(SCHED, #accountEmpty(ACCTTO), GCAP, GAVAIL, VALUE, #sizeLVals(RETS), intSizes(ARGS)) ... </k>
+    rule <k> #compute [ _, RETS::LValues = call _ at ACCTTO ( ARGS ) send VALUE , gaslimit GCAP, SCHED ] => Ccall(SCHED, #accountEmpty(ACCTTO), GCAP *Int Sgasdivisor < SCHED >, GAVAIL, VALUE, #sizeLVals(RETS), Ccallarg(SCHED, ARGS)) ... </k>
```
bmmoore: Shouldn't we divide by `Sgasdivisior` and round at the contract return? If we don't, then the exact fine-grained cost left over could affect whether the caller runs out of gas even when the different amounts would round to the same cost.

iele-gas.md
```diff
@@ -739,7 +753,7 @@ Note: These are all functions as the operator `#compute` has already loaded all
     rule Cgascap(SCHED, GCAP, GAVAIL, GEXTRA) => minInt(#allBut64th(GAVAIL -Int GEXTRA), GCAP) requires GAVAIL >=Int GEXTRA andBool notBool Gstaticcalldepth << SCHED >>
     rule Cgascap(SCHED, GCAP, GAVAIL, GEXTRA) => GCAP                                          requires GAVAIL <Int  GEXTRA orBool Gstaticcalldepth << SCHED >>
 
-    rule Cextra(SCHED, ISEMPTY, VALUE, RETS, ARGS) => Gcall < SCHED > +Int Cnew(SCHED, ISEMPTY, VALUE) +Int Cxfer(SCHED, VALUE) +Int Ccallmem(SCHED, RETS, ARGS)
+    rule Cextra(SCHED, ISEMPTY, VALUE, RETS, ARGS) => Gcall < SCHED > +Int Cnew(SCHED, ISEMPTY, VALUE) +Int Cxfer(SCHED, VALUE) +Int Gcallreg < SCHED > *Int (RETS +Int ARGS)
```
bmmoore: Why did this change to `Gcallreg` when `Ccallmem` was renamed to `Ccallarg` above?

iele-gas.md
```diff
@@ -828,22 +845,39 @@ Note: These are all functions as the operator `#compute` has already loaded all
       requires L1 >=Int L2
     
     rule Cdiv(SCHED, L1, L2) =>
-        Gdivword < SCHED > *Int L1 +Int //TODO(remove this term)
+        Gdivword < SCHED > *Int L1 +Int
+        Gdiv < SCHED >
+      requires notBool Gnewarith << SCHED >>
+      [owise]
+
+    rule Cdiv(SCHED, L1, L2) =>
```
bmmoore: Why doesn't the new div cost depend on the size of the numbers at all?

iele-gas.md
```diff
@@ -1012,16 +1066,19 @@ This schedule is used to execute the EVM VM tests, and contains minor variations
...
+    rule Sgasdivisor    < DEFAULT > => 1
 ```
 
 ### Albe Schedule
 
 This is the initial schedule of IELE.
 
 ```k
-    syntax Schedule ::= "ALBE"
+ // Albe
```
bmmoore: Where did `syntax Schedule ::= "ALBE"` go?

iele-gas.md
```diff
@@ -828,22 +845,39 @@ Note: These are all functions as the operator `#compute` has already loaded all
...
+
+    rule Cdiv(SCHED, L1, L2) =>
         Gdiv < SCHED >
+      requires Gnewarith << SCHED >>
       [owise]
 
     rule Cexp(SCHED, L1, W1, W2) =>
         Gexpkara < SCHED > *Int #overApproxKara(#adjustedBitLength(L1, W1) *Int W2 /Int 64) +Int
-        Gexpword < SCHED > *Int L1 +Int // TODO(adjustedBitLength)
+        Gexpword < SCHED > *Int L1 +Int
```
bmmoore: What is the explanation of this new formula?

iele-gas.md
dwightguth: It was moved to well-formedness.k because it's needed now by the well-formedness semantics.

iele-gas.md
```diff
@@ -543,10 +543,19 @@ Each of these operations pays a constant cost to look up information about an ac
 The cost to load a value into a register is simply the cost to copy its value.
 
 ```k
-    rule <k> #compute [ DEST = % SRC:Int, SCHED ] => Gcopy < SCHED > *Int intSize(getInt(REGS [ SRC ])) ... </k>
+    rule <k> #compute [ DEST = % SRC:Int, SCHED ] => Gmove < SCHED > *Int intSize(getInt(REGS [ SRC ])) ... </k>
```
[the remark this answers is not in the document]
dwightguth: no, we just deleted Gcopy and merged it with Gmove since they were the same value before.

iele-gas.md
dwightguth: Since Gnewmove is always true in new versions of IELE, I think that's fine.

iele-gas.md
dwightguth: Can you give me an example of your concern? I'm having trouble following what you see as the problem here.

iele-gas.md
dwightguth: Because we measured it and it doesn't depend on them in this case; this is the case when you divide a smaller number by a larger one and get zero.

dwightguth: whoops, looks like I messed up.

iele-gas.md
dwightguth: it's the same formula, but we measure L1 at bit-granularity. This gave us a more accurate fit when we were measuring cpu time.

iele-gas.md
dwightguth: ah, no, this is correct. note that this is only the owise case when L1 <Int L2

iele-gas.md
dwightguth: Ccallarg is the number of register arguments to the call. Gcallreg is the cost per regisetr.

iele-gas.md
bmmoore: Ah, the diff context didn't show the case above.

iele-gas.md
bmmoore: That covers `Cexp`, but what about the extra terms for `Cexpmod`.

iele-gas.md
bmmoore: If we are not trying to hide the fine-grained gas costs at inter-contract boundaries then this is fine.

iele.md
```diff
@@ -908,7 +899,7 @@ These operators make queries about the current execution state.
 -   `REG = call @iele.blockhash(N)` returns the hash of the block header of the Nth previous block.
 
 ```k
-    rule <k> #exec REG = call @iele.gas      ( .Ints )  => #load REG GAVAIL ... </k> <gas> GAVAIL </gas>
+    rule <k> #exec REG = call @iele.gas      ( .Ints )  => #load REG (GAVAIL up/Int Sgasdivisor < SCHED >) ... </k> <gas> GAVAIL </gas> <schedule> SCHED </schedule>
```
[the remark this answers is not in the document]
bmmoore: With it rounding up we could have `iele.gas` return an amount of gas remaining, and then trying to pass that much in a call could fail.

iele-gas.md
dwightguth: these were also based on improving the fit in our measurements.

PR #137, opened 2018-06-22
Corrections to install instructions.

marick:
Two changes:

1. `caskroom/version` no longer exists. The new cask is named `versions`. Also `caskroom` is deprecated in favor of `homebrew`.

2. I didn't realize I was building from source rather than setting up a runtime environment, so when I hit `git submodule`, I stumbled that was the first clue I was supposed to be in a cloned repo.


marick:
I have executed every line of the new instructions on my machine. 

It would be ideal to check the instructions on a pristine machine so as to reveal unsuspected dependencies. I haven't done that. However, the original instructions break on the first line, so these changes are certainly not a regression.

dwightguth:
Jenkins: add to whitelist

PR #141, opened 2018-07-11
update k submodule and add alias for unparsing

dwightguth:


dwightguth:
Jenkins: test this please

PR #162, opened 2018-11-21
Add KORE submodule and update Makefile

dfilaretti:
I added KORE as a submodule in IELE. Notice it's in the folder `tests/ci/kore`; I am not sure this is the right choice, but I tried to mimic K which is in `tests/ci/rv-k`. 

The Makefile has been updated to provide the following: 

   * `make haskell-deps`: builds the haskell backend 
   * `make haskell`: compiles IELE to KORE, to be used with the Haskell backend.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,6 +16,6 @@
 [submodule ".build/tangle"]
 	path = .build/tangle
 	url = https://github.com/ehildenb/pandoc-tangle
-[submodule "tests/ci/kore"]
-	path = tests/ci/kore
+[submodule ".build/kore"]
+	path = .build/kore
 	url = https://github.com/kframework/kore
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,11 +2,6 @@
 	path = tests/ethereum-tests
 	url = https://github.com/ethereum/tests.git
 	branch = develop
-[submodule "tests/ci/rv-k"]
-	path = tests/ci/rv-k
-	url = https://github.com/runtimeverification/k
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -16,3 +16,6 @@
 [submodule ".build/tangle"]
 	path = .build/tangle
 	url = https://github.com/ehildenb/pandoc-tangle
+[submodule "tests/ci/kore"]
```
dwightguth: Can we move both of these to .build?

Makefile
```diff
@@ -180,11 +187,20 @@ ocaml-deps:
 	opam switch 4.03.1+k
 	eval `opam config env` && opam install -y mlgmp zarith uuidm cryptokit secp256k1.0.3.2 bn128 hex ocaml-protoc rlp yojson ocp-ocamlres bisect_ppx
 
+haskell-deps: kore-deps
+		git submodule update --init -- $(KORE_SUBMODULE)
```
dwightguth: I don't want this to be a part of the Makefile. It should already be part of the rest of the install instructions.

.gitmodules
dfilaretti: You mean moving both `kore` and `rv-k`, right?

dfilaretti: Done. Those submodules are now in `.build/rv-k` and `.build/kore`. 
 
What was previously `.build/kore/iele-testing.kore` is now `.build/llvm/iele-testing.kore`, and the `Makefile` target has been renamed from `kore` to `llvm` accordingly.

PR #164, opened 2018-11-27
Run in haskell backend

dfilaretti:
Adding a simple script, similar to `vmtest`, that runs a IELE program on the Haskell backend. 
It's not fully functional/finished.

### DIFF
--- a/vmtest-haskell
+++ b/vmtest-haskell
@@ -4,10 +4,9 @@ KRUN_BINARY=".build/rv-k/k-distribution/target/release/k/bin/krun"
 KOMPILED_DEF=".build/haskell/"
 HASKELL_BACKEND_CMD=".build/kore/bin/kore-exec"
 
-cSCHEDULE='`DEFAULT_IELE-GAS`(.KList)'
-cMODE='`VMTESTS_IELE-CONSTANTS`(.KList)'
+cSCHEDULE=DEFAULT
+cMODE=VMTESTS
 
 PGM=$1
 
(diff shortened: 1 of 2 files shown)


vmtest-haskell
```diff
@@ -0,0 +1,12 @@
...
+KRUN_BINARY=".build/rv-k/k-distribution/target/release/k/bin/krun"
+KOMPILED_DEF=".build/haskell/"
+HASKELL_BACKEND_CMD=".build/kore/bin/kore-exec"
+
+cSCHEDULE=DEFAULT
+cMODE=VMTESTS
+
+PGM=$1
+
+$KRUN_BINARY --directory "$KOMPILED_DEF" -cSCHEDULE="$cSCHEDULE" -pSCHEDULE='printf %s' -cMODE="$cMODE" -pMODE='printf %s' "$PGM" --haskell-backend-command "$HASKELL_BACKEND_CMD"
```
dwightguth: setting the parser to printf %s means that you are expected to pass kast to the -c flag, but you're not here.

dfilaretti: Understood. I saw that used in KEVM, wasn't sure what it was for but for some reason I thought it was needed. Removed now! 

By the way, I used to get the same error regardless of what test I was trying to run, but now it is gone, so it was probably caused by those bad command line settings.

dfilaretti:
@dwightguth can I merge this or are there other changes that I need?

PR #170, opened 2019-01-18
update k submodule

dwightguth:


ehildenb:
@dwightguth make sure you agree with the modifications to the `k-deps` target.

Issue #189, opened 2020-10-20
Move IELE specific stuff out of blockchain-k-plugin


ehildenb:
Currently the blockchain-k-plugin has a bunch of stuff that is IELE specific, and should be moved to the IELE repository. This includes:

-   OCaml files for bindings.
-   Protobuf protocol for communicating with Mantis client.

PR #193, opened 2020-10-21
Bring in IELE specific files from plugin submodule

gtrepta:
fixes: https://github.com/runtimeverification/iele-semantics/issues/189

So far this moves the `vm` and `vm-c` subfolders from plugin into the base of the repo. This would render the files in the submodule dead and can be removed in a later pull request.

I suspect `plugin-c` has some files that can go as well, specifically `blockchain.cpp`, `world.{c,h}` and the `proto` folder. I would like to get some confirmation on that.


dwightguth:
Looks good, assuming you didn't modify any of the added files.

gtrepta:
> Looks good, assuming you didn't modify any of the added files.

They're unmodified. Do you know if moving the other files I mentioned in the description will be fine?

dwightguth:
Yes, those files can probably be moved because I believe kevm is not using them anymore.

ehildenb (approved):

PR #204, opened 2020-11-06
Fix iele-vm tests for Milestone 2

dwightguth:


ehildenb (approved):

ehildenb:
What's up with all the deleted files?

dwightguth:
previously we had a field in the protocol buffer that indicated whether to use the old or new gas schedule. This was inserted specifically to support the old testnet that had mined blocks prior to the fixing of a bunch of bugs in the gas schedule. IOG never added that corresponding field to their version of the protocol buffer, so we had two choices. Either get them to make that change on their end, or simply deprecate the old gas schedule since we don't intend to use it for anything anymore. The next step is to just delete it from the codebase, but I figured that could be a separate PR as a cleanup step.  But I had to delete the tests from this PR because the IELE VM doesn't know how to pick anything but the new gas schedule anymore, so those tests were failing.

PR #216, opened 2020-11-27
Ubuntu Package Releases

ehildenb:
Fixes: https://github.com/runtimeverification/iog-pm/issues/26


ehildenb:
Evidence that this works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/iele-semantics/detail/PR-216/40/pipeline

package/debian/control
```diff
@@ -0,0 +1,15 @@
...
+Maintainer: Everett Hildenbrandt <everett.hildenbrandt@runtimeverification.com>
+Build-Depends: clang-8 , cmake , kframework , libcrypto++-dev , libjemalloc-dev , libprocps-dev , libprotobuf-dev , libsecp256k1-dev , libssl-dev , opam , pkg-config , protobuf-compiler
+Standards-Version: 3.9.6
+Homepage: https://github.com/runtimeverification/iele-semantics
+
+Package: kiele
+Architecture: any
+Section: devel
+Priority: optional
+Depends: libcrypto++-dev , libjemalloc-dev , libmpfr-dev , libprotobuf-dev , libsecp256k1-dev
```
dwightguth, line 13: this doesn't look right. If you're not compiling C code, which I don't believe the user of the IELE package is, then you ought not to need the development version of these libraries.

package/debian/rules
```diff
@@ -0,0 +1,32 @@
...
+#export DEB_CFLAGS_MAINT_APPEND  = -Wall -pedantic
+# package maintainers to append LDFLAGS
+#export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
+
+%:
+	dh $@
+
+override_dh_auto_build:
+	git submodule update --init --recursive
+	make COVERAGE=k
```
dwightguth, line 21: this is wrong for a release; we don't want to build the coverage build except on CI.

package/debian/test-package.sh
```diff
@@ -0,0 +1,22 @@
...
+
+git clone 'https://github.com/runtimeverification/iele-semantics'
+cd iele-semantics
+git checkout "$KIELE_REVISION"
+git submodule update --init --recursive
+
+iele-vm 0 127.0.0.1 > port &
+sleep 3
+export PORT=$(cat port | awk -F ':' '{print $2}')
+make test -j`nproc` -k TEST_WELLFORMED=true
```
dwightguth, line 21: I don't understand why we're running all the tests except for the check-iele tests. why would we be skipping those?

dwightguth (changes requested):

package/debian/test-package.sh
ehildenb: Because the check-iele tests require access to `krun` and `kast`, which we are not assuming are installed on the system. So I just disable them and will add a card for properly packaging the `check-iele` semantics later.

dwightguth: Ah, that's reasonable.

ehildenb:
Evidence that it works; https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/iele-semantics/detail/PR-216/44/pipeline

dwightguth (approved):

PR #222, opened 2020-12-04
kiele runner script

ehildenb:
Fixes: runtimeverification/iog-pm#27

This PR does a few things:

-   Adds a runner wrapper script `kiele`, which subsumes all the functionality of the old runners `*test`.
-   Adds all the needed binaries to the install, and the `*-kompiled` directories, to have things like `kprint` able to work.
-   Splits the tests in the Jenkinsfile into logical parallel units.
-   Reworks the testing harness to use the runner script rather than individual test runners.


package/debian/test-package.sh
```diff
@@ -15,8 +15,12 @@ cd iele-semantics
 git checkout "$KIELE_REVISION"
 git submodule update --init --recursive
 
-iele-vm 0 127.0.0.1 > port &
+make test-vm -j4
+make test-iele -j4
+
+kiele vm --port 9001 &
```
dwightguth, line 21: I would really much rather we kept things with a dynamically allocated port rather than a statically allocated port. This is a step backwards because the build will crash if this port is already busy.

Makefile
```diff
@@ -88,86 +74,98 @@ protobuf_out := $(BUILD_DIR)/plugin-node/proto/msg.pb.cc
...
 
 # Tests
 # -----
 
+TEST          = kiele
+TEST_ASSEMBLE = ./assemble-iele-test
+TEST_BACKEND  = standalone
+TEST_MODE     = NORMAL
+TEST_SCHEDULE = DEFAULT
+TEST_PORT     = 10000
```
dwightguth, line 88: here is another case where we have taken a step backwards and picked a statically numbered port.

dwightguth (changes requested):
Aside from my inline comments, I would also like to note that we are no longer building coverage data on each run of CI, which is something that IOG is likely to ask us to do again at some point, so it would behoove us to make sure it is easy to add that back in. I'm not sure that that is true right now because if we were to build and run the tests with coverage enabled, that would then enable coverage in the release we deploy.

package/debian/test-package.sh
ehildenb: Not sure why the port would already be busy in the container on CI where this runs, but I can make the change.

But FYI, we've used statically determined ports for ages on teh Firefly build and that has never caused problems.

dwightguth: It probably doesn't matter if it's just CI, but it needs to be able to work with a dynamically allocated port locally at the very least.

ehildenb:
The `Makefile` alllows overriding the `TEST_PORT` in the standard way, I can make `test-package.sh` also allow overriding it if you like.

As far as the coverage build, I did not disable it or remove it, just did not do it by default. I can change that back if you ilke.

ehildenb:
@dwightguth 

-   You can specify the TEST_PORT on commandline for make in the standard way `TEST_PORT=....`
-   You can now also specify the `TEST_PORT` for `test-package.sh` now.
-   I don't think we need to enable coverage build by default.

dwightguth (approved):
This looks okay for now, just bear in mind we may eventually need to modify this so that the tests are run with coverage enabled but the package is built with coverage disabled.

PR #240, opened 2021-01-20
Refactoring for use as a submodule

gtrepta:
Fixes: https://github.com/runtimeverification/iog-pm/issues/53


gtrepta:
Some things that need to be done:

- `#initCoverage` and `#finishCoverage` are in place, but are not being used yet. I don't yet know where to insert them into the execution and how to pick the correct `<bytecodeCoverage>` cell. I have tried but so far have had no success.
- Actual coverage recording for each instruction such as visited, branch taken/not taken, exception thrown still need to be put in place. This is blocked on the above item.
- We should discuss how we want to index the instructions in a contract. Currently I am indexing them starting from 0 at the beginning of each contract declaration within a whole contract, with the reasoning being that if a contract does a `create` operation then the new account with the contract wont have instructions starting from a non-zero index. If we want to number them starting from zero globally in the contract then we will have to be careful of this.

Makefile
```diff
@@ -217,6 +220,7 @@ build-testnode : $(IELE_TEST_VM) $(IELE_TEST_CLIENT)
 KOMPILE=kompile
 
 KOMPILE_INCLUDE_OPTS := -ccopt -I -ccopt $(PLUGIN)/plugin-c -ccopt -I -ccopt $(PROTO) -ccopt -I -ccopt $(BUILD_DIR)/plugin-node -ccopt -I -ccopt $(LOCAL_INCLUDE)
+KOMPILE_INCLUDE_OPTS += -I $(IELE_DIR)
```
ehildenb, line 223: Why line split?

Also, maybe consider improving the first line here with `$(addprefix -ccopt , $(LLVM_KOMPILE_INCLUDES))`, and define the `LLVM_KOMPILE_INCLUDES` variable above.

kiele
```diff
@@ -137,6 +138,7 @@ while [[ $# -gt 0 ]]; do
         --module)     module="$2"               ; shift 2 ;;
         --definition) definition="$2"           ; shift 2 ;;
         --schedule)   schedule='`'$2'`(.KList)' ; shift 2 ;;
+        --depth)      depth=$2                  ; shift 2 ;;
```
ehildenb, line 141: If we're adding the `--depth` option, we need to make sure it's handled correctly by all cases of `interpret` (you've only handled the `standalone` case), and we should also make it handle correctly for `krun` option.

Makefile
gtrepta: The two lines represent two different classes of arguments, as well as keeping line length more sane looking. The addprefix option is a good idea though, should be useful even without the variable.

ehildenb (approved):

PR #244, opened 2021-02-11
Add KIELE script to Nix package

ttuegel:
Fixes runtimeverification/iog-pm#56.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -338,7 +338,10 @@ $(INSTALL_BIN)/%: $(IELE_BIN)/%
 $(INSTALL_LIB)/%: $(IELE_LIB)/%
 	install -D $< $@
 
-$(INSTALL_BIN)/kiele: $(patsubst %, $(INSTALL_LIB)/%, $(install_libs))
+$(INSTALL_BIN)/iele-interpreter: $(INSTALL_LIB)/standalone/iele-testing-kompiled/syntaxDefinition.kore
+
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/kore-json.py
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/version
 
 install: $(patsubst %, $(INSTALL_BIN)/%, $(install_bins))
 
--- a/Makefile
+++ b/Makefile
@@ -331,6 +331,9 @@ $(IELE_LIB)/version:
 $(IELE_LIB)/standalone/iele-testing-kompiled/%: $(BUILD_DIR)/standalone/iele-testing-kompiled/interpreter
 	install -D $(dir $<)$* $@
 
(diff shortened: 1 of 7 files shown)


kiele
```diff
@@ -12,11 +12,11 @@ export PATH="$INSTALL_BIN:$INSTALL_LIB:$PATH"
...
 check_k_install() {
-    which kore-exec &> /dev/null \
+    type -p kore-exec &> /dev/null \
         || fatal "Must have K installed! See https://github.com/kframework/k/releases."
-    which kast &> /dev/null \
+    type -p kast &> /dev/null \
         || fatal "Must have K installed! See https://github.com/kframework/k/releases."
-    which krun &> /dev/null \
+    type -p krun &> /dev/null \
         || fatal "Must have K installed! See https://github.com/kframework/k/releases."
```
ehildenb, line 20: What's the reasoning behind using `type` instead of `which`? Because maybe `which` is not installed and if they have `bash` then they have `type`?

Makefile
```diff
@@ -336,12 +343,24 @@ $(INSTALL_BIN)/%: $(IELE_BIN)/%
...
 	install -D $< $@
 
-install: $(patsubst %, $(INSTALL_BIN)/%, $(install_bins)) $(patsubst %, $(INSTALL_LIB)/%, $(install_libs))
+$(INSTALL_BIN)/iele-interpreter: $(INSTALL_LIB)/standalone/iele-testing-kompiled/syntaxDefinition.kore
+
+$(INSTALL_BIN)/iele-check: $(INSTALL_LIB)/check/well-formedness-kompiled/compiled.bin
+$(INSTALL_BIN)/iele-check: $(INSTALL_LIB)/check/well-formedness-kompiled/interpreter
+
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/kore-json.py
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/version
```
ehildenb, line 352: Did you have a chance to test that this works for:

-   When the well-formedness checker passes _and_ when it fails?
-   Same for the interpreter (exit code 0, no printing, exit code non-zero, printing configuration)?

No worries if not, we can make a follow-up PR to make all that work. Just wondering.

ehildenb (approved):

kiele
ttuegel: There's not really any notion of "installed" with Nix. I can make `kiele` depend on `which`, but `type` is built-in to Bash and does the same thing in this case. This is a standard tweak that we often have to make to scripts that are being packaged with Nix, but I have no problem pulling in the `which` package if you prefer.

Makefile
ttuegel: The Nix test suite runs `test-interactive`, `test-iele`, `test-wellformed`, and `test-illformed`. I don't know which of your points that covers.

PR #245, opened 2021-02-11
Update the MacOS build instructions

gtrepta:
fixes: https://github.com/runtimeverification/iog-pm/issues/52


Makefile
```diff
@@ -42,9 +49,10 @@ IELE_VM          := $(IELE_BIN)/iele-vm
 IELE_TEST_VM     := $(IELE_BIN)/iele-test-vm
 IELE_TEST_CLIENT := $(IELE_BIN)/iele-test-client
 
+SHELL=/bin/bash
```
ehildenb, line 52: Why is this one?

kiele
```diff
@@ -42,15 +42,19 @@ run_interpret() {
...
             ;;
         haskell)
             check_k_install
-            depth_switch=()
-            [[ "$depth" == "-1" ]] || depth_switch+=(--depth "$depth")
-            kore-exec "$backend_dir/$definition-kompiled/definition.kore" --pattern "$kast" --module "$module" --output "$output" "${depth_switch[@]}" "$@" || status="$?"
+            kore_cmd=(kore-exec "$backend_dir/$definition-kompiled/definition.kore" --pattern "$kast" --module "$module" --output "$output")
+            [[ "$depth" == "-1" ]] || kore_cmd+=(--depth "$depth")
+            kore_cmd+=("$@")
+            ${kore_cmd[@]} || status="$?"
```
ehildenb, line 57: Why do we have quotes around `${interpreter_cmd[@]}` and not around `${kore_cmd[@]}`? I seem to vaguely remember that this actually matters, but we should double-check.

Makefile
```diff
@@ -220,30 +256,30 @@ build-testnode : $(IELE_TEST_VM) $(IELE_TEST_CLIENT)
...
+$(BUILD_DIR)/check/well-formedness-kompiled/interpreter: $(checker_files) $(protobuf_out) $(libff_out) $(libsecp256k1_out) $(libcryptopp_out)
 	$(KOMPILE) --debug --main-module IELE-WELL-FORMEDNESS-STANDALONE --md-selector "(k & ! node) | standalone" \
 	                                --syntax-module IELE-SYNTAX well-formedness.md --directory $(BUILD_DIR)/check --hook-namespaces KRYPTO \
 	                                --backend llvm -ccopt $(protobuf_out) $(KOMPILE_CPP_OPTS) $(KOMPILE_INCLUDE_OPTS) $(KOMPILE_LINK_OPTS) -ccopt -g -ccopt -std=c++14 -ccopt -O2 $(KOMPILE_FLAGS)
 
 $(BUILD_DIR)/standalone/iele-testing-kompiled/interpreter: MD_SELECTOR="(k & ! node) | standalone"
 $(BUILD_DIR)/node/iele-testing-kompiled/interpreter: MD_SELECTOR="(k & ! standalone) | node"
 
-$(BUILD_DIR)/%/iele-testing-kompiled/interpreter: $(k_files) $(protobuf_out) $(libff_out)
+$(BUILD_DIR)/%/iele-testing-kompiled/interpreter: $(k_files) $(protobuf_out) $(libff_out) $(libsecp256k1_out) $(libcryptopp_out)
```
ehildenb, line 275: This seems like it would require that Ubuntu users, for example, now have to always build libsecp256k1 and libcryptopp from source even if they have it globally installed.

You may be able to put the definition of `libsecp256k1_out` in an `ifdef` block above to make it so that it's empty if we're not on MacOS.

kiele
gtrepta: I just forgot to put quotes around kore_cmd.

Makefile
gtrepta: For some reason, if you don't set SHELL then setting PATH in a Makefile doesn't work on Mac. I only needed to do this so I could run the tests after building, so I can remove it if you prefer.

https://stackoverflow.com/a/25506676

ehildenb: Include a link to this post in a comment above this then.

ehildenb (approved):
Please add the comment about `SHELL` in the Makefile, then add automerge

PR #251, opened 2021-03-04
Single `test` rule in Makefile for all tests

gtrepta:
fixes: https://github.com/runtimeverification/iog-pm/issues/61

### DIFF
--- a/kiele
+++ b/kiele
@@ -81,7 +81,7 @@ run_check() {
 }
 
 run_vm() {
-    iele-vm "$kiele_port" "$kiele_host"
+    exec iele-vm "$kiele_port" "$kiele_host"
 }
 
 # Main
--- a/tests/node-test.sh
+++ b/tests/node-test.sh
@@ -1,7 +1,21 @@
 #!/bin/bash
 
-kiele vm --port 9001 &
+port=9001
+
+args=()
+while [[ $# -gt 0 ]]; do
+    arg="$1"
+    case $arg in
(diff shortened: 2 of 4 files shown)


tests/node-test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+kiele vm --port 9001 &
+sleep 3
+make test-iele-node  "$@" TEST_PORT=9001
+make test-bad-packet "$@" TEST_PORT=9001
+kill %kiele
```
ehildenb, line 7: Allow passing in `TEST_PORT` as an argument, and have it be a `Makefile` variable as well, so that users can call `make test-node TEST_PORT=9003` if they need to.

kiele
```diff
@@ -81,7 +81,7 @@ run_check() {
 }
 
 run_vm() {
-    iele-vm "$kiele_port" "$kiele_host"
+    exec iele-vm "$kiele_port" "$kiele_host"
```
ehildenb, line 84: What's this change for?

gtrepta: The original test script in the Jenkinsfile had an error, it recorded the pid of kiele so it could kill it at the end of the script. But the kiele script was invoking iele-vm as a new process, so the kill command wasn't killing the vm. The kiele script does nothing after launching the iele-vm, so I made this change which makes the kill command actually work.

ehildenb (approved):

PR #252, opened 2021-03-10
Fix haskell backend install

gtrepta:
fixes: https://github.com/runtimeverification/iog-pm/issues/62

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -412,6 +412,8 @@ $(INSTALL_BIN)/iele-check: $(INSTALL_LIB)/check/well-formedness-kompiled/interpr
 
 $(INSTALL_BIN)/kiele: $(INSTALL_LIB)/kore-json.py
 $(INSTALL_BIN)/kiele: $(INSTALL_LIB)/version
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/haskell/iele-testing-kompiled/definition.kore
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/haskell/iele-testing-kompiled/syntaxDefinition.kore
 
 install: $(patsubst %, $(INSTALL_BIN)/%, $(install_bins))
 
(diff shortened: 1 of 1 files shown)


ttuegel:
@gtrepta Yes. The insight is: I need those Makefile targets you removed.

gtrepta:
Are those just the `$(INSTALL_BIN)/...` targets, or also that one `$(IELE_LIB)/check/well-formedness-kompiled/%` target? Will the modified rules be ok?

Makefile
```diff
@@ -390,30 +394,30 @@ $(IELE_LIB)/standalone/iele-testing-kompiled/%: $(BUILD_DIR)/standalone/iele-tes
...
-
 $(INSTALL_BIN)/iele-interpreter: $(INSTALL_LIB)/standalone/iele-testing-kompiled/syntaxDefinition.kore
 
 $(INSTALL_BIN)/iele-check: $(INSTALL_LIB)/check/well-formedness-kompiled/compiled.bin
 $(INSTALL_BIN)/iele-check: $(INSTALL_LIB)/check/well-formedness-kompiled/interpreter
 
 $(INSTALL_BIN)/kiele: $(INSTALL_LIB)/kore-json.py
 $(INSTALL_BIN)/kiele: $(INSTALL_LIB)/version
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/haskell/iele-testing-kompiled/definition.kore
+$(INSTALL_BIN)/kiele: $(INSTALL_LIB)/haskell/iele-testing-kompiled/syntaxDefinition.kore
```
ttuegel, line 409: @gtrepta The problem is here: the Nix package doesn't include the Haskell backend, but you have made it a prerequisite for installing `kiele`.

gtrepta: I'm trying to get these files included in the installation because they are needed by the `kiele interpret --backend haskell` command.

gtrepta: @ttuegel So the nix package for K doesn't include the haskell backend? How hard would it be to add that in?

ttuegel: The Nix package for K includes the Haskell backend. The Nix package for IELE does not. This should not be harder than pulling it into the package, but I will need you to find a way that those files don't become dependencies of the recipe to install `kiele`, because those will be built at a different time.

ehildenb (approved):

PR #266, opened 2021-04-12
Coverage collection for Firefly

gtrepta:
fixes: https://github.com/runtimeverification/iog-pm/issues/55


vm/c/vm.cpp
```diff
@@ -243,6 +258,21 @@ CallResult run_transaction(CallContext ctx) {
...
+    extractedCoverages* extractedCoverage = (extractedCoverages *)evaluateFunctionSymbol(tag4, arr);
+    auto bytecodecoverages = k_to_coverage(&extractedCoverage->bytecodecoverages->data);
+    for (bytecodecoverage_cell *bytecodecoverage : bytecodecoverages) {
+      auto coveragemsg = result.add_coveragedata();
+      std::string codehash = of_z_width(32, bytecodecoverage->bytecodeHash->data);
+      std::string coveragebytes = std::string(bytecodecoverage->coverageData->data->data, len(bytecodecoverage->coverageData->data));
+      coveragemsg->set_codehash(codehash);
+      coveragemsg->set_coverage(coveragebytes);
+    }
+  }
```
ehildenb, line 274: @dwightguth this one block I don't feel very comfortable reviewing, could you or @theo25 take a look and make sure it makes sense?

ehildenb:
This looks good to me overall, but I just want to double-check that the larger chunk of LLVM-backend code marked above makes sense to either @dwightguth or @theo25 .

vm/c/vm.cpp
dwightguth: This looks fine but it looks like you're leaving the bytecode field of the protobuf message blank? Is that deliberate?

vm/ieleClientUtils.ml
```diff
@@ -261,7 +261,7 @@ let output_framed out_chan encoder v =
 
 let send addr ctx =
   let chans = Unix.open_connection addr in
-  let hello = {version="1.1";config=Iele_config} in
+  let hello = {version="1.2";config=Iele_config} in
```
dwightguth, line 264: Same here as above.

vm/c/vm.cpp
gtrepta: At the moment yes, it's deliberate, although I guess it wouldn't hurt to add it in. I included the bytecode field in the spec since it might be needed by the Firefly frontend, but we haven't got that far yet.

ehildenb (approved):

PR #267, opened 2021-04-15
Run Test workflow on macOS

### LINKED ISSUE #268: Build KIELE Nix package on macOS
The LLVM backend is working now, but another dependency fails to build on macOS.

ttuegel:
We must run the Nix workflow on macOS to populate the binary cache.

Fixes #268.


ttuegel:
This would fix runtimeverification/solidity#175.

andreiburdusa:
Why did you add `autoconf`, `automake` and `libtool` but removed them later?

ttuegel:
> Why did you add `autoconf`, `automake` and `libtool` but removed them later?

They were dependencies for one part of the Makefile, but then I realized that part needed to be disabled.

andreiburdusa (approved):

PR #290, opened 2021-06-11
Generate static coverage reports through the firefly command line

shd101wyy:
Fixes: https://github.com/runtimeverification/web-common/issues/235

Steps:

* [x] Parse the artifacts on the client side (Translate the back-end Haskell code to Python)
* [x] Build a single HTML template file
* [x] Build a script (maybe in Python3?) to parse the artifacts that user would originally firefly upload . We inject the parsed result into the HTML remplate file to generate a single HTML report file.
* [x] Add simple tests


kiele
```diff
@@ -99,6 +99,19 @@ run_vm() {
...
 
+run_generate_report() {
+    if [[ -f $(dirname $(which kiele))/static-report.html ]]; then
+        report_template=$(dirname $(which kiele))/static-report.html
+    else
+        report_template=$INSTALL_LIB/static-report.html
+    fi
+
+    report_json=${1:-"./report.json"}
+    output_path=${2:-""}
```
gtrepta, line 110: It's probably a better idea to require the positional parameter instead of have a fallback value, you have it documented that way as well (`<...>` is required, `[...]` is optional). What happens when the user tries to give just the output file here?

How about using the `run_file` variable declared at the bottom of the script instead of declaring `report_json`?

kiele
```diff
@@ -181,16 +198,17 @@ cmode='Lbl'$mode'{}()'
 cschedule='Lbl'$schedule'{}()'
 cenablecoverage='\dv{SortBool{}}("'$enablecoverage'")'
 
-if [[ "$run_command" != 'vm' ]]; then
+if [[ "$run_command" != 'vm' && "$run_command" != "generate-report" ]]; then
```
gtrepta, line 201:
```suggestion
if [[ "$run_command" != 'vm' ]]; then
```
With this `run_file` is already made for you so you don't have to declare `report_json`. It looks like you would need to handle `-` as a special case, just reassigning `run_file` to `/dev/stdin` should work.

gtrepta (commented):
Looks good! I made some suggestions for the `kiele` script.

You also have the option to add and use a `-o|--output` parameter for the output html file. My PR for generating the `report.json` file is going to add that to the script regardless.

After you see my comments and make whatever changes you want to make, re-request my review and I can approve it.

gtrepta (approved):

PR #291, opened 2021-06-16
fix for bswap instruction

theo25:
This PR removes an artificial limitation on the semantics of the `bswap` instruction, namely the instruction now can be applied to a negative integer argument.


ehildenb (approved):

ehildenb:
I added the breaking change, DO NOT MERGE until that commit is reverted. This should fail tests anyway now though.

PR #292, opened 2021-06-17
tests/node-test.sh: fix shebang and options

ehildenb:


theo25 (approved):

ehildenb:
I added the breaking change, DO NOT MERGE until that is reverted. It should fail tests now anyway.

ehildenb:
Confirmed that we catch breakages correctly now

PR #300, opened 2021-08-09
code cell is populated before initialization code runs

theo25:


iele.md
```diff
@@ -1445,11 +1446,6 @@ For each `call*` operation, we make a corresponding call to `#call` and a state-
          <mode> EXECMODE </mode>
          <gas> GAVAIL </gas>
          <output> _ => ACCT , .Ints </output>
-         <account>
-           <acctID> ACCT </acctID>
-           <code> _ => CODE </code>
-           ...
-         </account>
 
     rule <k> #exception STATUS ~> #finishCodeDeposit _ _ REG _ NEW:Bool => #popCallStack ~> #popWorldState ~> #popSubstate ~> #if NEW #then STATUS #else #registerDelta(REG, 1) ~> #load REG STATUS #fi ... </k>
```
ehildenb, line 1450: I guess the `CODE` variable is no longer used by `#finishCodeDeposit` (and maybe thus not used by the hing that calls `#finishCodeDeposit`, so maybe we can remove that variable argument from those things?

Also, did you check whether #mkCall needs similar treatment?

theo25: You are right, I pushed a commit that removes the unused parameter.

About the second question, `#mkCall` should be fine as the code cell of the callee should have  already been initialized at the creation of the callee contract. So, even when the cell is empty in the configuration, `#getCode` (the builtin that queries the blockchain client) should be able to fetch the code and populate the cell. That was not the case with `#mkCreate` before this change, because the contract in question was just being created and the code cell was empty even in the state of the blockchain client. Hence, `#getCode` in that case would return `#emptyCode`.

iele.md
```diff
@@ -1419,7 +1420,7 @@ For each `call*` operation, we make a corresponding call to `#call` and a state-
 
     syntax KItem ::= "#codeDeposit" Int Int Contract LValue LValue Bool
                    | "#mkCodeDeposit" Int Int Contract LValue LValue Bool
-                   | "#finishCodeDeposit" Int Contract LValue LValue Bool
+                   | "#finishCodeDeposit" Int LValue LValue Bool
```
ehildenb, line 1423: It looks to me like none of #codeDeposit, #mkCodeDeposit need the CODE parameter either actually.

dwightguth:
Question: does the intialization of the code cell still get reverted with this change if the constructor raises an exception?

iele.md
theo25: Sounds right, I have removed it from those too.

ehildenb (approved):

PR #306, opened 2021-08-27
Build cleanups

ehildenb:
These cleanups are largely needed to get KIELE building with the newest K releases, but should also be standalone.


ehildenb:
Must be merged with #289

PR #317, opened 2021-10-01
Fix execution getting stuck when indexing contracts

gtrepta:


anvacaru (approved):
This seems to fix the issue and now I am able to deploy contracts successfully 👍

PR #318, opened 2021-10-12
Add test runner python module

anvacaru:
## Writing tests
 - First, the module `pyiele` needs to be imported.
 - The default values for the client `port`, `passphrase`, `spending-key`, `gas-limit` and other parameters are defined in the [config.py](https://github.com/anvacaru/iele-test-runner/blob/main/pyiele/config.py) file. 
 - A contract can be deployed using `deploy_contract(walletId, sender, bytecode)`.
 - A contract function can be called using `run_function(f_name, f_args, walletId, sender, to)`.
 - Any rpc call can be made using `send(rpc)`.
 - RPC calls are defined in [rpc.py](https://github.com/anvacaru/iele-test-runner/blob/main/pyiele/rpc.py).

## Running tests
```shell
>$ python3 test_erc20.sol
```


Makefile
```diff
@@ -473,6 +486,7 @@ install-check: $(patsubst %, $(DESTDIR)$(INSTALL_LIB)/%, $(iele_check_libs))
 
 install-kiele: $(patsubst $(IELE_BIN)/%, $(DESTDIR)$(INSTALL_BIN)/%, $(IELE_RUNNER))
 install-kiele: $(patsubst %, $(DESTDIR)$(INSTALL_LIB)/%, $(kiele_files))
+install-kiele: $(patsubst %, $(DESTDIR)$(INSTALL_LIB)/%, $(pyiele_files))
```
ehildenb, line 489:
```suggestion
install-kiele: $(patsubst $(IELE_BIN)/%, $(DESTDIR)$(INSTALL_BIN)/%, $(IELE_RUNNER)) \
               $(patsubst %, $(DESTDIR)$(INSTALL_LIB)/%, $(kiele_files))             \
               $(patsubst %, $(DESTDIR)$(INSTALL_LIB)/%, $(pyiele_files))
```

kiele
```diff
@@ -111,20 +116,20 @@ run_vm() {
     "${cmd[@]}"
 }
 
+run_coverage() {
+    notif "Generating coverage json from $run_file and client on port $kiele_port"
+    if ! nc -z "$kiele_host" "$kiele_port"; then
+        notif "No listener found on port $kiele_port, so there wont be any coverage data!"
```
ehildenb, line 122: Should this be `fatal` instead?

pyiele/__main__.py
```diff
@@ -0,0 +1,169 @@
...
+    return parser
+
+
+if __name__ == '__main__':
+    parser = make_parser()
+    args = parser.parse_args()
+
+    if args.command == 'compile':
+        config.contracts_dir=args.contracts_dir
+        config.target_directory=args.build_dir
```
ehildenb, line 73:
```suggestion
        config.contracts_dir    = args.contracts_dir
        config.target_directory = args.build_dir
```

pyiele/__main__.py
```diff
@@ -0,0 +1,169 @@
...
+                contract_calls, constructor_info = blackbox_test_single_contract(contract, available_contract)
+                all_calls.extend(contract_calls)
+            if len(constructor_info) == 0:
+                all_constructors[solidity_file] = []
+            else:
+                all_constructors[solidity_file] = constructor_info[0]["input"]
+
+        # Write founded tests to blackbox-random.json
+        with open('blackbox-random.json', 'w') as outfile:
+            json.dump(all_calls, outfile)
```
ehildenb, line 143: How is this file used? Is it incorporated into the web report too?

pyiele/config.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+
+class config:
+  port = 8342
+  passphrase = "walletNotSecure"
+  master_key = "m-test-shl-mk1ypc6mjyustmuma5vf2kvxm2aml2clfewfmk7mvkuk8344vdxzgwseytxyg"
+  gas_price = "0x0"
+  gas_limit = "0x166f5777"
+  target_directory = "contract_artifacts"
```
ehildenb, line 9: Is this file passed as an argument, or is it hardcoded right now? I guess it would be nice if users could pass it as an argument?

ehildenb:
@anvacaru and @gtrepta can we add some tests of this functionality as well? Just a single Solidity file -> generated report files, and diff the files should be enough.

Or we can just make sure the tool runs smoothly, and not record the output, as an initial test. But it probably would be better to start with too strong of testing (with the output files) and weaken it as it gets annoying.

pyiele/__main__.py
anvacaru: AFAIK, this file is not used in the web report currently.

pyiele/config.py
anvacaru: Most of the values used in the config class can be overwritten by values passed by the user as command line arguments.

kiele
anvacaru: Yes, it should. I'll change that.

kiele
```diff
@@ -47,6 +48,10 @@ run_assemble() {
     iele-assemble "$run_file" "$@"
 }
 
+run_pyiele() {
+    python3 -m pyiele "$@"
+}
```
ehildenb, line 53:
```suggestion
run_pyiele() {
    python3 -m pyiele --host "${host}" --port "${port}" --output "${output_path}" "$@"
}
```

kiele
```diff
@@ -111,20 +116,21 @@ run_vm() {
     "${cmd[@]}"
 }
 
+run_coverage() {
+
+    notif "Generating coverage json from and client on port $kiele_port"
+    if ! nc -z "$kiele_host" "$kiele_port"; then
+        fatal "No listener found on port $kiele_port, so there wont be any coverage data!"
+    fi
+    run_pyiele coverage --port "$kiele_port" --output "$output_path" $@
```
ehildenb, line 125:
```suggestion
    run_pyiele coverage "$@"
```

kiele
```diff
@@ -141,6 +147,47 @@ run_upload_report() {
...
+    else
+      run_pyiele test --port "$kiele_port" "$@"
+    fi
+}
+
+run_blackbox() {
+    if [[ $# -gt 0 ]] && [ "$1" == "--help" ]; then
+      run_pyiele test --help
+      exit
+    fi
```
ehildenb, line 184:
```suggestion
```

kiele
```diff
@@ -141,6 +147,47 @@ run_upload_report() {
...
+    if ! nc -z "$kiele_host" "$kiele_port"; then
+        fatal "No listener found on port $kiele_port"
+    fi
```
ehildenb, line 187: Factor out a `check_listener` function in the Utilities section at the top, and use that.

kiele
gtrepta: This will have to be
```
python3 -m pyiele "$@" --host "${host}" --port "${port}" --output "${output_path}"
```
Because the command `{blackbox,test,compile,coverage}` is a positional parameter.

gtrepta: On second look, this might not work at all. None of these parameters are in each of the commands.

kiele
```diff
@@ -141,6 +147,47 @@ run_upload_report() {
...
+        else
+            run_pyiele test --file "$run_file" --port "$kiele_port" "$@"
+        fi
+      else
+        run_pyiele test --port "$kiele_port" "$@"
+      fi
+    else
+      run_pyiele test --port "$kiele_port" "$@"
+    fi
+}
```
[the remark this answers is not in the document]
gtrepta: My suggestion:

```suggestion
run_testrunner() {
    local run_file cmd
    if [[ $# -gt 0 ]] && [ "$1" == "--help" ]; then
      run_pyiele test --help
      exit
    fi
    if ! nc -z "$kiele_host" "$kiele_port"; then
        fatal "No listener found on port $kiele_port"
    fi

    cmd=(run_pyiele test --port "$kiele_port" --host "$kiele_host")

    if [[ $# -gt 0 ]] && ([[ -f "$1" ]] || [[ -d "$1" ]]); then
      run_file="$1"; shift
      if [[ "$run_file" == *.py ]]; then
          cmd=( python3 "$run_file" )
      else
          cmd+=( --file "$run_file" )
      fi
    fi

    "${cmd[@]}" "$@"
}
```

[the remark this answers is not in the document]
ehildenb: The `--help` option should be handled once globally at the top, and if we want it to call the pyiele help there as well, then it should do that.

But we shouldn't duplicate all this stuff about calling `help` everywhere.

kiele
anvacaru: I was able to set the `run_pyiele()` function to Everett's suggestion by moving the `port`,`host`, and `output` parameters to the main parser (initially, they were subparser parameters).

kiele
anvacaru: Done.

kiele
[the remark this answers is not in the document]
anvacaru: I agree with @ehildenb. Initially, I added explicit `help` calls because the `check_listener` would prevent the `help` message from being displayed if the midnight client wasn't active. For now I just removed all explicit calls.

kiele
```diff
@@ -141,6 +150,34 @@ run_upload_report() {
...
+        else
+            run_pyiele test --file "$run_file" "$@"
+        fi
+      else
+        run_pyiele test "$@"
+      fi
+    else
+      run_pyiele test "$@"
+    fi
+}
```
gtrepta, line 174: You can still collapse those if blocks with this, otherwise it looks good to me.
```suggestion
run_testrunner() {
    local run_file cmd
    check_listener
    cmd=(run_pyiele test)
    if [[ $# -gt 0 ]] && ([[ -f "$1" ]] || [[ -d "$1" ]]); then
      run_file="$1"; shift
      if [[ "$run_file" == *.py ]]; then
        cmd=(python3 "$run_file")
      else
        cmd+=(--file "$run_file")
      fi
    fi
    "${cmd[@]}" "$@"
}
```

gtrepta (approved):

ehildenb (approved):

PR #330, opened 2021-11-10
INSTALL.md: update installation directions

SchmErik:
In addition to the opam command changes, I've fixed broken links as well as file paths that I've ran into while installing the iele-semantics

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -49,14 +49,14 @@ Copy all the files in the tarball into place:
 
 ```sh
 mkdir -p ~/.local/bin ~/.local/lib
-cp -r kiele-X.Y.Z-bin/bin/* ~/.local/bin/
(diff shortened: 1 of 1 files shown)


theo25 (approved):
This looks fine to me.

Edit: Make sure to make the fix the Dwight suggested before you merge though.

INSTALL.md
```diff
@@ -195,7 +196,7 @@ These commands build and install KIELE:
 git clone https://github.com/runtimeverification/iele-semantics.git
 cd iele-semantics
 git submodule update --init --recursive
-k-configure-opam
+opam init --yes
```
dwightguth, line 199: these are interactive instructions, so we should not pass --yes in case users wish to configure their environment differently than the default.

SchmErik: thanks for pointing that out

PR #346, opened 2022-04-18
Fix CI and K dependency

Baltoli:
This PR points the package script to the correct URL for the Github release page, and updates the K dependency to the newest version (the old version currently being used doesn't get picked up by the releases API).


Baltoli:
The repo is too far behind upstream K for this to be an easy fix; closing.

